PR #79, opened 2021-06-30
Fixed issues with path during deploy to Object Storage from Windows

nanov94:


src/utils/index.ts
```diff
@@ -0,0 +1,7 @@
+export * from './logger';
```
nanov94, line 1: Only `path` new one, other were moved to separate files

src/utils/logger.ts
```diff
@@ -0,0 +1,27 @@
...
+import {ArgvService} from '../services';
+
+function writeLog(msg: string) {
+    const {quiet} = ArgvService.getConfig();
+
+    if (quiet) {
+        return;
+    }
+
+    console.log(msg);
```
vovcyan, line 11: Looks like forgotten

src/utils/path.ts
```diff
@@ -0,0 +1,20 @@
+import {Platforms} from '../constants';
+
+const windowsSlash = '\\';
+const linuxSlash = '/';
```
vovcyan, line 4: I think isn't a good way to use name of platforms for slashes, cause Mac is not a linux, but also using normal slashes.

What about just `slash` and `backslash`?

src/constants.ts
```diff
@@ -22,6 +22,10 @@ export enum Lang {
     EN = 'en',
 }
 
+export enum Platforms {
+    WINDOWS = 'win32',
```
vovcyan, line 26: Let's add mac and linux here for consistency

src/utils/index.ts
vovcyan: Gotcha

src/utils/logger.ts
nanov94: I changed nothing in this code 
https://github.com/yandex-cloud/yfm-docs/blob/master/src/utils.ts#L99-L107

src/constants.ts
nanov94: Yep, I agree with your concern, however we don't use them at the moment. 

Ok, I will add it

src/utils/logger.ts
vovcyan: Sorry, missed that

vovcyan (approved):
LGTM

PR #82, opened 2021-07-02
Fixing bug with warns about custom authors

nanov94:


src/services/authors.ts
```diff
@@ -20,14 +20,23 @@ async function updateAuthorMetadataString(defaultMetadata = '', vcsConnector?: V
...
 
-    return null;
+    try {
+        JSON.parse(author);
+        return author.replace(/"/g, '\'');
+    } catch {
+        const user = await vcsConnector.getUserByLogin(author);
+
+        if (user) {
+            return JSON.stringify(user).replace(/"/g, '\'');
```
vovcyan, line 35: Should we move this logic to separate method with described comments?
For now it looks like secret knowledge

vovcyan (approved):

PR #85, opened 2021-07-13
Vladvlad/add presets system section to page meta

gorgeousvlad:


src/resolvers/md2md.ts
```diff
@@ -12,14 +12,17 @@ import {getContentWithUpdatedMetadata} from '../services/metadata';
...
-        content = await getContentWithUpdatedMetadata(metadata, content);
-    }
+    const vars = {
+        ...PresetService.get(dirname(inputPath)),
+        ...configVars,
+    };
+    const content = await getContentWithUpdatedMetadata(
+        readFileSync(resolvedInputPath, 'utf8'),
+        metadata,
+        vars.__system,
```
vovcyan, line 24: Why'd not just `system`?

src/services/metadata.ts
```diff
@@ -2,8 +2,18 @@ import {VCSConnector} from '../vcs-connector/connector-models';
 import {MetaDataOptions} from '../models';
 import {getAuthorDetails, updateAuthorMetadataString} from './authors';
 import {getFileContributorsMetadata, getFileContributorsString} from './contributors';
-
-async function getContentWithUpdatedMetadata(options: MetaDataOptions, fileContent: string): Promise<string> {
+import {replaceDoubleToSingleQuotes} from '../utils';
+import {isObject} from './utils';
+
+async function
+getContentWithUpdatedMetadata(
```
vovcyan, line 9: `async function getContentWithUpdatedMetadata(` ?

vovcyan (approved):

src/resolvers/md2md.ts
gorgeousvlad: Eugene says it should be __system

PR #93, opened 2021-09-23
Linter

yndx-birman:


yndx-birman:
Need to merge first https://github.com/yandex-cloud/yfm-transform/pull/84

src/index.ts
```diff
@@ -143,6 +149,7 @@ async function main() {
         const outputBundlePath: string = join(outputFolderPath, BUNDLE_FOLDER);
         const pathToConfig = _yargs.argv.config || join(_yargs.argv.input, '.yfm');
         const pathToRedirects = join(_yargs.argv.input, REDIRECTS_FILENAME);
+        const pathToLintConfig = join(_yargs.argv.input, LINT_CONFIG_FILENAME);
```
vovcyan, line 152: Do we really need separate file for lint config? Can we combine it with `.yfm` ?

src/constants.ts
```diff
@@ -26,6 +26,7 @@ export const TMP_OUTPUT_FOLDER = '.tmp_output';
 export const MAIN_TIMER_ID = 'Build time';
 export const SINGLE_PAGE_FOLDER = '_single_page';
 export const REDIRECTS_FILENAME = 'redirects.yaml';
+export const LINT_CONFIG_FILENAME = 'yfmlint.yaml';
```
vovcyan, line 29: .yfm-lint ?

src/index.ts
```diff
@@ -143,6 +149,7 @@ async function main() {
         const outputBundlePath: string = join(outputFolderPath, BUNDLE_FOLDER);
         const pathToConfig = _yargs.argv.config || join(_yargs.argv.input, '.yfm');
```
vovcyan, line 150: Let's move `.yfm` name into constants too

src/constants.ts
yndx-birman: Why don't we want to use the yaml extension if `.yfm` includes yaml syntax?  `.yfm` includes yaml syntax and it's not highlighted in IDEs as yaml

vovcyan: That's true, but this approach separates hidden service files started with dot from documentation related files with `.yaml` extension.

PR #144, opened 2022-04-06
feat: add support text array in TextItem

Feverqwe:


src/services/utils.ts
```diff
@@ -68,7 +68,11 @@ export function filterTextItems(
         const useItem = shouldProcessItem(item, vars, options);
 
         if (useItem) {
-            result.push(item.text);
+            if (Array.isArray(item.text)) {
+                result.push(...item.text);
+            } else {
+                result.push(item.text);
+            }
```
Lunatic174, line 75: I suppose we can shorten it
```suggestion
            result.push(Array.isArray(item.text)  ? (...item.text) : item.text)
```

Lunatic174 (approved):

src/services/utils.ts
Feverqwe: Typescript don't understand this syntax, so, maybe better leave it as is

PR #146, opened 2022-04-08
DOCSTOOLS-1103: await publishing files to S3

yndx-birman:


src/steps/publishFilesToS3.ts
```diff
@@ -42,10 +42,10 @@ export function publishFilesToS3(): void {
...
         logger.upload(pathToFile);
 
-        s3Client.upload(params, (error) => {
+        return s3Client.upload(params, (error) => {
             if (error) {
                 throw error;
             }
-        });
-    }
+        }).promise();
```
vovcyan, line 49: Do we really need callback to handler error after using promise?

vovcyan (approved):

src/steps/publishFilesToS3.ts
vovcyan: I think that we can keep only promise in this case

yndx-birman: We don't need it, it's useless, I'll remove it

PR #149, opened 2022-05-05
feat: add textItems support in title, meta.title, nav.title, and toc.title

Feverqwe:


src/services/leading.ts
```diff
@@ -25,11 +25,34 @@ function filterFile(path: string) {
...
+
+        if (parsedIndex.meta?.title) {
+            parsedIndex.meta.title = firstFilterTextItems(
+                parsedIndex.meta.title,
+                combinedVars,
+                {resolveConditions: true},
+            );
+        }
+
+        if (parsedIndex.nav) {
```
yndx-birman, line 48: Should the title property be checked the way you checked it above?

yndx-birman (approved):

src/services/leading.ts
Feverqwe: In nav model title required. It may be optional?

PR #155, opened 2022-06-09
DOCSTOOLS-1190: use title from meta data or the first header as page …

yndx-birman:
…title

### DIFF
--- a/src/resolvers/md2html.ts
+++ b/src/resolvers/md2html.ts
@@ -105,5 +105,6 @@ function MdFileTransformer(content: string, transformOptions: FileTransformOptio
         path,
         assetsPublicPath,
         getVarsPerFile: getVarsPerRelativeFile,
+        extractTitle: true,
     });
 }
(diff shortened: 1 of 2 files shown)


src/resolvers/md2html.ts
```diff
@@ -105,5 +105,6 @@ function MdFileTransformer(content: string, transformOptions: FileTransformOptio
         path,
         assetsPublicPath,
         getVarsPerFile: getVarsPerRelativeFile,
+        extractTitle: true,
```
martyanovandrey, line 108: not used anywhere ?

martyanovandrey (approved):

src/resolvers/md2html.ts
martyanovandrey: nvm, its for @doc-tools/transform

yndx-birman: The `extractTitle` option is used to return the first first-level header as the title of the entire document
Without this parameter the title is not defined

PR #166, opened 2022-08-22
fix: Fix linters warnings

burashka:


src/resolvers/md2md.ts
```diff
@@ -34,10 +34,6 @@ export async function resolveMd2Md(options: ResolveMd2MdOptions): Promise<string
         copyFile,
     });
 
-    if (singlePage) {
```
burashka, line 37: Looks like never use

yndx-birman (approved):

PR #179, opened 2022-09-09
feat: support pdf file generation

yndx-birman:
Merge first [this](https://github.com/yandex-cloud/yfm-transform/pull/171/commits) and then update transform dependency here

How to try:
`npm i -g @doc-tools/docs@1.31.0-beta.0`
`yfm --input $inputPath --output $outputPath --pdf-file`


package.json
```diff
@@ -35,6 +35,7 @@
     "js-yaml": "4.1.0",
     "mime-types": "2.1.34",
     "node-html-parser": "^5.3.3",
+    "puppeteer": "^17.1.1",
```
burashka, line 38: It has overweight. Are you sure that need in default package? It was separate package in previous life.

yndx-birman:
I got this error running on Ubuntu

Screen Shot 2022-11-11 at 9 01 42 AM

Found explanation:
https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md#running-puppeteer-in-docker

>The bundled Chromium that Puppeteer installs is missing the necessary shared library dependencies.
To fix, you'll need to install the missing dependencies

PR #180, opened 2022-09-12
feat: add conditions and substitutions in title, desc

Feverqwe:
Add conditions and substitutions support in title, description, meta.title, links[*].title, links[*].description and toc.title

### DIFF
--- a/src/services/utils.ts
+++ b/src/services/utils.ts
@@ -117,12 +117,12 @@ function shouldProcessItem<T extends Filter>(item: T, vars: Record<string, strin
 }
 
 export function liquidFields(fields: undefined | string | string[], vars: Record<string, unknown>, path: string) {
-    if (!fields) {
-        return fields;
-    }
-
     if (typeof fields === 'string') {
         return liquidField(fields, vars, path);
+    }
+
+    if (!Array.isArray(fields)) {
+        return fields;
     }
 
     return fields.map((item) => {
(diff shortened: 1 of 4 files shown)


src/services/utils.ts
```diff
@@ -114,6 +116,32 @@ function shouldProcessItem<T extends Filter>(item: T, vars: Record<string, strin
...
 }
 
+export function liquidFields(fields: undefined | string | string[], vars: Record<string, unknown>, path: string) {
+    if (!fields) {
+        return fields;
+    }
+    if (typeof fields === 'string') {
+        return liquidField(fields, vars, path);
+    }
+    return fields.map((item) => {
```
burashka, line 126: I see typings, but may be add check for array? Looks like it is yaml and we don't validate his structure

src/services/utils.ts
```diff
@@ -114,6 +116,32 @@ function shouldProcessItem<T extends Filter>(item: T, vars: Record<string, strin
     return useItem;
 }
 
+export function liquidFields(fields: undefined | string | string[], vars: Record<string, unknown>, path: string) {
+    if (!fields) {
+        return fields;
+    }
+    if (typeof fields === 'string') {
+        return liquidField(fields, vars, path);
+    }
```
burashka, line 125: Pls, add empty rows between blocks: ifs and return

src/services/utils.ts
```diff
@@ -114,6 +116,32 @@ function shouldProcessItem<T extends Filter>(item: T, vars: Record<string, strin
...
+    return fields.map((item) => {
+        return liquidField(item, vars, path);
+    });
+}
+
+export function liquidField(input: string, vars: Record<string, unknown>, path: string) {
+    const {applyPresets, resolveConditions} = ArgvService.getConfig();
+    if (!applyPresets && !resolveConditions) {
```
burashka, line 133: Pls, add empty row between declaration and if

burashka (approved):

PR #198, opened 2022-10-28
fix(services/tocs): fix `when` processing order

3y3:


3y3:
Related with
https://st.yandex-team.ru/DOCSTOOLS-1478

yndx-birman (approved):

PR #226, opened 2022-12-20
fix(includers/openapi): openapi generator skips path without tags

amosov-f:
Creating section for paths without tags.
Result example for merchant openapi:
Снимок экрана 2022-12-20 в 20 56 07

### DIFF
--- a/src/services/includers/batteries/openapi/index.ts
+++ b/src/services/includers/batteries/openapi/index.ts
@@ -10,8 +10,9 @@ import {
     YfmToc,
     IncluderFnParams,
     IncluderFnOutput,
+    IncluderFnOutputElement,
 } from '../../../../models';
-import {Info, Tag} from './types';
+import {Endpoint, Info} from './types';
 
 async function generateContent(params: IncluderFnParams): IncluderFnOutput {
     const {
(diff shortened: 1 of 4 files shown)


src/services/includers/batteries/openapi/index.ts
```diff
@@ -58,6 +58,13 @@ async function generateContent(params: IncluderFnParams): IncluderFnOutput {
         });
     });
 
+    for (const endpoint of spec.endpoints) {
```
moki, line 61: refactor into handleEndpointRender or smth like that

moki: handle both render of the endpoints grouped by tags and endpoints without tags

src/services/includers/batteries/openapi/index.ts
```diff
@@ -115,13 +122,20 @@ async function generateTocs(params: IncluderFnParams): IncluderFnOutput {
...
                 href: path,
-                name: endpoint.summary ?? endpoint.id,
+                name: sectionName(endpoint),
             } as YfmToc);
         });
 
         result.content.items.push(section);
     });
 
+    for (const endpoint of endpoints) {
```
moki, line 132: same as in content generation

src/services/includers/batteries/openapi/parsers.ts
```diff
@@ -153,11 +157,14 @@ function paths(spec: OpenapiSpec, tagsByID: Map<string, Tag>): Map<string, Tag>
                 endpoints: old.endpoints.concat(parsedEndpoint),
             });
         }
+        if (!tags.length) {
```
moki, line 160: separate with empty line for loop and if statement

moki (changes requested):
left some refactoring comments, the rest lgtm

src/services/includers/batteries/openapi/parsers.ts
amosov-f: fixed

src/services/includers/batteries/openapi/index.ts
amosov-f: fixed

src/services/includers/batteries/openapi/index.ts
amosov-f: fixed

moki (approved):

PR #236, opened 2022-12-29
fix(interceptors): intercept links without extension

Feverqwe:
Fix handle link


moki:
we currently have big pending release, i don't want potentally breaking changes to be merged right now.

https://github.com/yandex-cloud/yfm-docs/pull/230

also use scope for the commit messages `feat(scope): commit message`

yndx-birman (approved):

PR #242, opened 2023-01-09
fix(includers/openapi): add description in tag name

zorox112-practicum:


moki:
we support 'x-navtitle' field inside openapis endpoint to specify pretty tag names like this

therefore one able to specify pretty(display) tag name for the endpoint to belong to.

```
      x-navtitle:
      - Departments
```

moki:
we support 'x-navtitle' field inside openapis endpoint to specify pretty tag names like this

therefore one able to specify pretty(display) tag name for the endpoint to belong to.

```
      x-navtitle:
      - Departments
```

PR #272, opened 2023-01-26
fix(includers/openapi): default value in query

amosov-f:
fix undefiened & default values

Снимок экрана 2023-01-26 в 16 42 32

### DIFF
--- a/src/services/includers/batteries/openapi/generators/endpoint.ts
+++ b/src/services/includers/batteries/openapi/generators/endpoint.ts
@@ -22,6 +22,7 @@ import {
 } from '../types';
 import stringify from 'json-stringify-safe';
 import {prepareTableRowData, prepareSampleObject, tableFromSchema, tableParameterName} from './traverse';
+import {concatNewLine} from '../../common';
 
 function endpoint(allRefs: Refs, data: Endpoint) {
     // try to remember, which tables we are already printed on page
@@ -84,12 +85,15 @@ function parameters(params?: Parameters) {
 
 function parameterRow(param: Parameter) {
     const row = prepareTableRowData({}, param.schema, param.name);
-    let description = param.description;
+    let description = param.description ?? '';
     if (row.description.length) {
-        description += `<br>${row.description}`;
+        description = concatNewLine(description, row.description);
     }
-    if (param.example) {
-        description += `<br>Example: \`${param.example}\``;
+    if (param.example !== undefined) {
+        description = concatNewLine(description, `Example: \`${param.example}\``);
+    }
+    if (param.default !== undefined) {
+        description = concatNewLine(description, `Default: \`${param.default}\``);
     }
     return [tableParameterName(param.name, param.required), row.type, description];
 }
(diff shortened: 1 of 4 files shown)


src/services/includers/batteries/openapi/generators/endpoint.ts
```diff
@@ -84,12 +84,15 @@ function parameters(params?: Parameters) {
...
     const row = prepareTableRowData({}, param.name, param.schema);
-    let description = param.description;
+    let description = param.description ?? '';
     if (row.description.length) {
-        description += `<br>${row.description}`;
+        description = concatNewLine(description, row.description);
     }
-    if (param.example) {
-        description += `<br>Example: \`${param.example}\``;
+    if (param.example !== undefined) {
```
moki, line 91: why not `param.example != null` why should we render:

```
`Example: \`${null}\``
```

src/services/includers/batteries/openapi/generators/endpoint.ts
```diff
@@ -84,12 +84,15 @@ function parameters(params?: Parameters) {
...
+        description = concatNewLine(description, `Example: \`${param.example}\``);
+    }
+    if (param.default !== undefined) {
```
moki, line 94: why not `param.default != null` why should we render:

```
`Example: \`${null}\``
```

src/services/includers/batteries/openapi/types.ts
```diff
@@ -100,8 +100,9 @@ export type Parameter = {
     name: string;
     in: In;
     required: boolean;
-    description: string;
-    example: string | number;
+    description?: string;
+    example?: string | number | boolean;
```
moki, line 104: would love separate type for that, that will signal about what kind of data we support

src/services/includers/batteries/openapi/types.ts
```diff
@@ -100,8 +100,9 @@ export type Parameter = {
...
+    default?: string | number | boolean;
```
moki, line 105: same here

moki (changes requested):

src/services/includers/batteries/openapi/generators/traverse.ts
```diff
@@ -197,6 +197,6 @@ function merge(value: JSONSchema6 | boolean): JSONSchema6 | undefined {
     return value;
 }
 
-function concatNewLine(prefix: string, suffix: string) {
+export function concatNewLine(prefix: string, suffix: string) {
```
moki, line 200: if its used somewhere else factor it out to `common` module

src/services/includers/batteries/openapi/generators/endpoint.ts
amosov-f: null is valid default value

src/services/includers/batteries/openapi/generators/endpoint.ts
amosov-f: null is valid default value

src/services/includers/batteries/openapi/types.ts
amosov-f: fixed

src/services/includers/batteries/openapi/types.ts
amosov-f: fixed

src/services/includers/batteries/openapi/generators/traverse.ts
amosov-f: fixed

3y3 (approved):

moki (approved):

PR #273, opened 2023-01-26
Fix some linter warnings

3y3:


moki:
@3y3 message should prolly be `chore: fix linter warnings`

PR #300, opened 2023-03-01
Fixes ui

zorox112-practicum:
replace strong to red *
add possible reasons
add ClipboardButton response body

Screenshot 2023-03-01 at 15 22 16
Screenshot 2023-03-01 at 15 22 50

Screenshot 2023-03-01 at 15 26 46

### DIFF
--- a/src/services/includers/batteries/openapi/generators/traverse.ts
+++ b/src/services/includers/batteries/openapi/generators/traverse.ts
@@ -3,6 +3,7 @@ import {JSONSchema6} from 'json-schema';
 import {table} from './common';
 import slugify from 'slugify';
 import {concatNewLine} from '../../common';
+import {openapiBlock} from './constants';
 
 type TableRow = [string, string, string];
 
@@ -11,7 +12,7 @@ function anchor(ref: string) {
 }
 
 export function tableParameterName(key: string, required?: boolean) {
-    return required ? `<strong>${key}*</strong>` : key;
+    return required ? `${key}<span class="${openapiBlock('required')}">*</span>` : key;
 }
 
 export function tableFromSchema(allRefs: Refs, schema: JSONSchema6): {content: string; tableRefs: string[]} {
--- a/src/services/includers/batteries/openapi/plugin/public/components/Error.tsx
+++ b/src/services/includers/batteries/openapi/plugin/public/components/Error.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import {Text, Card} from '@gravity-ui/uikit';
 
-import {Text as TextEnum, yfmSandbox} from '../../constants';
+import {Text as TextEnum, yfmSandbox, possibleReasonsFailToFetch} from '../../constants';
 import {ErrorState} from '../types';
 import {Column} from './';
 
@@ -15,10 +15,18 @@ export const Error = ({message}: ErrorState) => {
             className={yfmSandbox('card')}
         >
             <Text
-                variant="body-3"
+                variant="code-2"
                 className={yfmSandbox('card-text')}
             >
-                {message}
+                {message} <br/>
+                Possible Reasons:
+                <ul className={yfmSandbox('possible-reasons')}>
+                    {possibleReasonsFailToFetch.map((item, index) => (
+                        <li key={index}>
+                            {item}
+                        </li>
+                    ))}
+                </ul>
             </Text>
         </Card>
     </Column>;
--- a/src/app/components/App/App.tsx
+++ b/src/app/components/App/App.tsx
@@ -15,6 +15,7 @@ import {getDocSettings, withSavingSetting, updateRootClassName} from '../../util
 import '../../interceptors/leading-page-links';
 
 import {Runtime as OpenapiSandbox} from '../../../services/includers/batteries/openapi/plugin/public';
+import '../../../services/includers/batteries/openapi/generators/style.scss';
 
 import '@doc-tools/components/styles/themes.scss';
 import '@doc-tools/components/styles/default.scss';
--- a/src/services/includers/batteries/openapi/generators/constants.ts
+++ b/src/services/includers/batteries/openapi/generators/constants.ts
@@ -1,0 +1,3 @@
+import block from 'bem-cn-lite';
+
+export const openapiBlock = block('openapi');
--- a/src/services/includers/batteries/openapi/generators/endpoint.ts
+++ b/src/services/includers/batteries/openapi/generators/endpoint.ts
@@ -24,6 +24,7 @@ import {
 import stringify from 'json-stringify-safe';
 import {prepareTableRowData, prepareSampleObject, tableFromSchema, tableParameterName} from './traverse';
 import {concatNewLine} from '../../common';
+import {openapiBlock} from './constants';
 
 function endpoint(allRefs: Refs, data: Endpoint, sandboxPlugin: {host?: string} | undefined) {
     // try to remember, which tables we are already printed on page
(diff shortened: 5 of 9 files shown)


src/services/includers/batteries/openapi/generators/traverse.ts
```diff
@@ -228,9 +228,19 @@ function merge(value: OpenJSONSchemaDefinition): OpenJSONSchema {
             properties[k] = v;
         }
     }
-    return {type: 'object', description, properties};
+    return {type: 'object', description, properties, allOf: value.allOf};
```
zorox112-practicum, line 231: allOf use for find ref

src/services/includers/batteries/openapi/generators/traverse.ts
```diff
@@ -228,9 +228,19 @@ function merge(value: OpenJSONSchemaDefinition): OpenJSONSchema {
...
 }
+
+function getType(value: JSONSchema6): JSONSchema6['type'] {
+    if (value.type) {
+        return value.type;
+    }
+    if (value.enum) {
+        return 'string';
+    }
+    return value.type;
```
moki, line 245: this whole function might as well be:

```
if (value.enum) {
  return 'string';
}

return value.type
```

or

```
return value.enum ?? value.type;
```

moki (changes requested):

src/services/includers/batteries/openapi/generators/traverse.ts
```diff
@@ -228,9 +228,19 @@ function merge(value: OpenJSONSchemaDefinition): OpenJSONSchema {
...
     }
-    return {type: 'object', description, properties};
+    return {type: 'object', description, properties, allOf: value.allOf};
 }
 
 function isRequired(key: string, value: JSONSchema6): boolean {
     return value.required?.includes(key) ?? false;
 }
+
+function getType(value: JSONSchema6): JSONSchema6['type'] {
```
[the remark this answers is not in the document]
zorox112-practicum: "any" in the example is a string, not a type, it seems that the remark is not relevant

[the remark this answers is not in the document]
zorox112-practicum: `getType  -> inferType` fixed

src/services/includers/batteries/openapi/generators/traverse.ts
zorox112-practicum: enum has the type "JSONSchema6Type[] | undefined" and it has possible values for enum, not their type. The purpose of the function is to set the string value "string" if the type was not specified in the specification

```
 export type JSONSchema6Type =
    | string 
    | number
    | boolean
    | JSONSchema6Object
    | JSONSchema6Array
    | null;
```

zorox112-practicum: This function should output the value of the "type" field if it is set explicitly, substitute "type: 'string'" if it is an enum and return the original type if it is not an enum, I can write in the last return like this:
`return undefined`

It seems that in the current version it is clearer that the function will return the original type than hardcoding `undefined`

moki: > This function should output the value of the "type" field if it is set explicitly, substitute "type: 'string'" if it is an enum and return the original type if it is not an enum, I can write in the last return like this: `return undefined`
> 
> It seems that in the current version it is clearer that the function will return the original type than hardcoding `undefined`

is there the case of:
type being set explicitly(to value other than 'string') and enum being provided for one entity?

zorox112-practicum: > type being set explicitly(to value other than 'string')
yes

> enum being provided for one entity
I don't understand this part of the question

moki: > > type being set explicitly(to value other than 'string')
> 
> yes
> 
> > enum being provided for one entity
> 
> I don't understand this part of the question

do we have a case when type is being set to value other that string and enum is also provided **at the same time**

zorox112-practicum: > do we have a case when type is being set to value other that string and enum is also provided at the same time

yes.
example#1
```
        name:
          type: integer
          enum:
            - 1
            - 2
```

Generated type:
```
export const PetNameEnum = {
    NUMBER_1: 1,
    NUMBER_2: 2
} as const;
```
example#2:
```
        name:
          type: string
          enum:
            - 1
            - 2
```
Generated type:
```
export const PetNameEnum = {
    _1: '1',
    _2: '2'
} as const;
```


In the OPEN API specification, enum is always an array.
https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.0/schema.json#L404

I fixed the type and made the logic, if type was not passed, then type is taken from enum[0]

src/services/includers/batteries/openapi/generators/traverse.ts
```diff
@@ -11,7 +11,7 @@ function anchor(ref: string) {
 }
 
 export function tableParameterName(key: string, required?: boolean) {
-    return required ? `<strong>${key}*</strong>` : key;
+    return required ? `${key}<span style="color: var(--yc-color-text-danger);">*</span>` : key;
```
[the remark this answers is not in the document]
zorox112-practicum: fixed:
https://github.com/yandex-cloud/yfm-docs/pull/300/commits/e9461d1ecfb177d91ae7c2b13641ca5c6ac44cd3

moki:
@zorox112-practicum i can approve ui changes as they are rn, and we will move on with this PR.

if you want to finish implementation of json6schema normalization, cherry pick commits that are about it onto another separate branch -> make PR then we will continue discussion there.

don't forget to cleanup commit history tho.

zorox112-practicum:
@moki ^

zorox112-practicum:
@moki 
this is the final version, sorry ^

moki:
@zorox112-practicum 

cleanup commit history:

rebase squash:

f0ec15e8f04c51b60b807c31399f92a71af783a9 into dc051c2c3c41718ce2fce9da74d091475b8ee6e6
c78258ee796acf33d6f405e8ac9d74635825ad97 into 88e27944370734a056d210d5da0b1dea42fd06c5

give meaningful name to each commit to have clear changelogs

describe functionality that you fix/add concisely in header and optionally go in depth inside the commit body.

commit messages:

f7bde2781f1ffb5e40650ba4c947b6eeb8dd6314 i see that you set data attribute for the link element why tho i have no idea, **describe what problem you are fixing with it**

dc051c2c3c41718ce2fce9da74d091475b8ee6e6 same goes for the **replacing strong with red**

6adb8e000483d85cc759bb5d013d14f4562c2dad better commit message would be `display request failure debug information`

src/services/includers/batteries/openapi/plugin/public/components/Error.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import {Text, Card} from '@gravity-ui/uikit';
 
-import {Text as TextEnum, yfmSandbox} from '../../constants';
+import {Text as TextEnum, yfmSandbox, possibleReasonsFailToFetch} from '../../constants';
```
moki, line 4: i've missed the point we started declaring constants in a camelCase

you wouldn't have to do `as TextEnum` if usual convention of having constants in uppercase SNAKE_CASE were followed.

about naming `FETCH_FAILURES` would be clear enough to describe constant values and more concise.

moki (approved):

PR #302, opened 2023-03-07
Add default type for enum

zorox112-practicum:
I made a function for calculating the enum type:
- if there is a type, we take it
- if there is no type, check that it is an enum and take its type

### DIFF
--- a/src/services/includers/batteries/openapi/constants.ts
+++ b/src/services/includers/batteries/openapi/constants.ts
@@ -20,6 +20,8 @@ const SPEC_RENDER_MODE_HIDDEN = 'hidden';
 const SPEC_SECTION_NAME = 'Specification';
 const SPEC_SECTION_TYPE = 'Open API';
 
+const SUPPORTED_ENUM_TYPES = ['string', 'number'] as const;
+
 export {
     TAG_NAMES_FIELD,
     BLOCK,
@@ -42,6 +44,7 @@ export {
     SPEC_RENDER_MODE_HIDDEN,
     SPEC_SECTION_NAME,
     SPEC_SECTION_TYPE,
+    SUPPORTED_ENUM_TYPES,
 };
 
 export default {
@@ -66,4 +69,5 @@ export default {
     SPEC_RENDER_MODE_HIDDEN,
     SPEC_SECTION_NAME,
     SPEC_SECTION_TYPE,
+    SUPPORTED_ENUM_TYPES,
 };
--- a/src/services/includers/batteries/openapi/generators/traverse.ts
+++ b/src/services/includers/batteries/openapi/generators/traverse.ts
@@ -1,8 +1,9 @@
-import {Refs} from '../types';
+import {JsType, Refs, SupportedEnumType} from '../types';
 import {JSONSchema6} from 'json-schema';
 import {table} from './common';
 import slugify from 'slugify';
 import {concatNewLine} from '../../common';
+import {SUPPORTED_ENUM_TYPES} from '../constants';
 
 type TableRow = [string, string, string];
 
@@ -20,7 +21,7 @@ export function tableFromSchema(allRefs: Refs, schema: JSONSchema6): {content: s
         const description = prepareComplexDescription('', schema);
         const content = table([
             ['Type', 'Description'],
-            [schema.type, description],
+            [inferType(schema), description],
         ]);
         return {content, tableRefs: []};
     }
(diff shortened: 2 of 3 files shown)


src/services/includers/batteries/openapi/generators/traverse.ts
```diff
@@ -108,7 +108,7 @@ function findRef(allRefs: Refs, value: JSONSchema6): string | undefined {
     }
     return undefined;
 }
-type OpenJSONSchema = JSONSchema6 & {example?: any};
+type OpenJSONSchema = Omit<JSONSchema6, 'enum'> & {example?: any; enum?: Array<any>};
```
moki, line 111: if you want to overwrite types, then construct appropriate type, exlcuding unsupported values(using exclude)

call it smth meaningful like Supported/ImplementedJSONSchema6EnumType

src/services/includers/batteries/openapi/generators/traverse.ts
```diff
@@ -234,3 +234,18 @@ function merge(value: OpenJSONSchemaDefinition): OpenJSONSchema {
 function isRequired(key: string, value: JSONSchema6): boolean {
     return value.required?.includes(key) ?? false;
 }
+
+function inferType(value: OpenJSONSchema): JSONSchema6['type'] {
+    if (value.type) {
+        return value.type;
+    }
+    if (value.enum && Array.isArray(value.enum)) {
```
moki, line 242: factor it out into separate function

`isEnumOfImplementedType`/`isEnumOfSupportedType` checking if value is one of the supported types

if so return that enum type, otherwise throwing

src/services/includers/batteries/openapi/generators/traverse.ts
```diff
@@ -234,3 +234,18 @@ function merge(value: OpenJSONSchemaDefinition): OpenJSONSchema {
 function isRequired(key: string, value: JSONSchema6): boolean {
     return value.required?.includes(key) ?? false;
 }
+
+function inferType(value: OpenJSONSchema): JSONSchema6['type'] {
```
moki, line 238: as far as i understand `inferType` might return `undefined` yet i can't find you handling that scenario

moki (changes requested):

src/services/includers/batteries/openapi/generators/traverse.ts
zorox112-practicum: This PR is about adding a type calculation for enum, this function does not handle undefined.

src/services/includers/batteries/openapi/generators/traverse.ts
zorox112-practicum: fixed

src/services/includers/batteries/openapi/generators/traverse.ts
zorox112-practicum: sorry, this is deprecated code.

deleted this code

src/services/includers/batteries/openapi/constants.ts
```diff
@@ -20,6 +20,8 @@ const SPEC_RENDER_MODE_HIDDEN = 'hidden';
 const SPEC_SECTION_NAME = 'Specification';
 const SPEC_SECTION_TYPE = 'Open API';
 
+const SUPPORTED_ENUM_TYPES = ['string', 'number'] as const;
```
zorox112-practicum, line 23: Please keep an array here for convenient type generation.

src/services/includers/batteries/openapi/generators/traverse.ts
```diff
@@ -234,3 +235,21 @@ function merge(value: OpenJSONSchemaDefinition): OpenJSONSchema {
...
+        if (isSupportedEnumType(enumType)) {
+            return enumType;
+        }
+        throw new Error('Not supported type enum');
+    }
+    return value.type;
+}
+
+function isSupportedEnumType(enumType: JsType): enumType is SupportedEnumType {
+    return SUPPORTED_ENUM_TYPES.some((type) => enumType === type);
```
zorox112-practicum, line 254: The includes method does not allow using TS

src/services/includers/batteries/openapi/generators/traverse.ts
moki: > This PR is about adding a type calculation for enum

1. you are adding function to infer type from value.

2. this function has undefined as one of its return values

3. either handle function returning undefined when you call it or throw error instead of returning undefined inside of the function.

zorox112-practicum: fixed

moki (approved):

PR #329, opened 2023-04-18
Add mermaid support

3y3:


esbuild.sh
```diff
@@ -0,0 +1,12 @@
+esbuild src/index.ts --outfile=build/index.js \
```
moki, line 1: why is it shell script instead of build script in js?

moki (changes requested):

moki:
why do we have three different contexts inside one PR?

- custom resources injection
- new plugin integration
- esbuild POC

moki (approved):

PR #397, opened 2023-06-08
fix: Resolve diplodoc/client bundle

3y3:


moki:
tested locally, it does fix problem with client app resolution

moki (approved):

PR #441, opened 2023-06-23
fix: Add missed styles in static render mode

3y3:


src/steps/processPages.ts
```diff
@@ -149,7 +149,7 @@ async function saveSinglePages(outputBundlePath: string) {
     }
 }
 
-function savePageResultForSinglePage(pageProps: ResolveMd2HTMLResult, pathData: PathData): void {
+function savePageResultForSinglePage(pageProps: DocInnerProps, pathData: PathData): void {
```
moki, line 152: the fact that now we have package that we depend upon dictating interface instead of complying with interface that could be defined in the main package is mind boggling to me.

moki (approved):

PR #465, opened 2023-06-29
fix: asyncify mapLimit callback, authorByGitEmail map added to reduce api request to GitHub

niktverd:


Feverqwe (approved):
Looks good

src/vcs-connector/github.ts
```diff
@@ -26,8 +26,10 @@ import {
 import {addSlashPrefix, logger} from '../utils';
 import {validateConnectorFields} from './connector-validator';
 
+const ignoreAuthor = process.env.IGNORE_AUTHOR;
```
moki, line 29: why are you taking ignoreAuthor out of env variables instead of config file?

moki: yeah, i think since this is not smth like credentials(auth tokens etc), we should specify it inside the config file.

putting smth like this in env variables "hides" it.

imagine someone would need "ignoreAuthor" parameter in other completely different context, overriding this one without even knowing it exists.

moki (changes requested):

src/vcs-connector/github.ts
niktverd: @Feverqwe asked to change this point

Feverqwe: @moki it looks like internal case, so we don't add it in flags.

moki: > @moki it looks like internal case, so we don't add it in flags.

read in my other comment, why its a bad idea.

niktverd: I got your point. I will return it back soon

moki (approved):
LGTM

PR #504, opened 2023-07-17
feat: ignore contributors if an email includes string from the --ignore-author parameter

niktverd:


Feverqwe (approved):
Looks good

src/vcs-connector/github.ts
```diff
@@ -124,6 +126,10 @@ async function matchContributionsForEachPath(repoLogs: string[], httpClientByTok
         const userData = dataArray[0];
         const [email, hashCommit] = userData.split(', ');
 
+        if (email.includes(ignoreAuthor)) {
```
Feverqwe, line 129: ignoreAuthor should not be empty

Feverqwe (approved):

moki (approved):

moki (changes requested):
I am just gonna point out that `email.includes(ignoreAuthor)` might exclude more than one originally intended.

let's say that we have user `gma` that we want to ignore.

the check `email.includes(ignoreAuthor)` will exclude all of the emails with the domain `@gmail.com`.

so, yeah... doesn't sound like a good filtering to me.

@3y3 wanna share you opinion on this?

@niktverd correct me if i misunderstood, but i just noticed this and it's like, well... really bad.

src/vcs-connector/github.ts
niktverd: fixed

3y3 (approved):

src/vcs-connector/github.ts
```diff
@@ -327,11 +331,29 @@ async function getAuthorForPath(path: string) {
...
-    if (ignoreAuthor && email.includes(ignoreAuthor)) {
+    if (shouldAuthorBeIgnored(email)) {
         return null;
     }
 
     return {hashCommit, normalizePath, email};
 }
 
+function shouldAuthorBeIgnored(email: string) {
+    const {ignoreAuthor} = ArgvService.getConfig();
```
moki, line 342: if `ignoreAuthor` becomes becomes collection of patterns that will be matched against, i think the name should reflect it.

how about calling it `ignoreAuthorPatterns`?

src/vcs-connector/github.ts
```diff
@@ -327,11 +331,29 @@ async function getAuthorForPath(path: string) {
...
+
+    if (!ignoreAuthor || typeof ignoreAuthor !== 'string') {
+        return false;
+    }
+
+    const emailPatterns = ignoreAuthor.split(',');
```
moki, line 348: correct me if i am wrong, but glob expressions can contain commas(`,`), therefore this will lead to splitting some of the expressions containing them, leading to undefined behaviour.

moki (changes requested):

moki (approved):
lgtm

src/vcs-connector/github.ts
niktverd: Good point. Done

src/vcs-connector/github.ts
niktverd: Good point, I have rearranged input option and made it an array

3y3 (approved):

PR #505, opened 2023-07-17
feat: change a log level while a request to github failed because a commit not found

niktverd:
…ot found

### DIFF
--- a/src/vcs-connector/client/github.ts
+++ b/src/vcs-connector/client/github.ts
@@ -8,6 +8,11 @@ import {
     GithubUserDTO,
 } from '../connector-models';
 import {validateConnectorFields} from '../connector-validator';
+import {logger} from '../../utils';
+
+interface ResponseError extends Error {
+    status: number;
+}
 
 async function getRepoUser(octokit: Octokit, username: string): Promise<GithubUserDTO | null> {
     try {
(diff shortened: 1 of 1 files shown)


src/vcs-connector/client/github.ts
```diff
@@ -1,5 +1,6 @@
 import log from '@doc-tools/transform/lib/log';
 import {Octokit} from '@octokit/core';
+import {GraphqlResponseError} from '@octokit/graphql';
```
3y3, line 3: This dependency in not installed in project

src/vcs-connector/client/github.ts
```diff
@@ -41,7 +42,14 @@ async function getRepoCommitByHash(httpClientByToken: Octokit, hashCommit: strin
 
         return commit.data;
     } catch (error) {
+        if ((error as GraphqlResponseError<{message: string}>)?.response?.data?.message?.includes(
```
3y3, line 45: Is there more precise class for `instanceof` check?

3y3 (changes requested):

src/vcs-connector/client/github.ts
niktverd: Removed it and replaced with custom class

src/vcs-connector/client/github.ts
niktverd: Replaced it with custom class. Error Type in @octokit made confused way

src/vcs-connector/client/github.ts
```diff
@@ -8,6 +8,17 @@ import {
     GithubUserDTO,
 } from '../connector-models';
 import {validateConnectorFields} from '../connector-validator';
+import {logger} from '../../utils';
+
+class ResponseError extends Error {
```
3y3, line 13: I don't see where this class will be instantiated.
Should it be a Type?

3y3 (changes requested):

3y3 (approved):

PR #524, opened 2023-07-31
Update @doc-tools/transform and add @diplodoc/tabs-extension

vsesh:


src/constants.ts
```diff
@@ -19,6 +19,7 @@ const term = require('@doc-tools/transform/lib/plugins/term');
 const changelog = require('@doc-tools/transform/lib/plugins/changelog');
 const mermaid = require('@diplodoc/mermaid-extension');
 const openapi = require('@diplodoc/openapi-extension');
+const tabs = require('@diplodoc/tabs-extension');
```
vsesh, line 22: Despite the fact that I replaced the tabs plugin in the `yfm-tranform` module, we still have to use the "@diplodoc/tabs-extension" module here directly because the plugin from `yfm-tranform` doesn't generate artifacts.

https://github.com/yandex-cloud/yfm-transform/pull/257/files#r1275362976

vsesh: But we don't need include the `@diplodoc/tabs-extension` module's CSS and client JS

martyanovandrey (approved):

Feverqwe (approved):

moki (approved):

PR #558, opened 2023-09-18
Add fs cache [NOTICKET]

Feverqwe:


Feverqwe:
Waiting https://github.com/diplodoc-platform/transform/pull/306

PR #595, opened 2023-11-17
fix(contributors): Force delete contributors branch

Feverqwe:


Feverqwe:
Found way to avoid `the branch '%s' is not fully merged` error.

PR #608, opened 2023-11-22
fix(deps): react-sortable-hoc

moki:


moki:
this PR contains empty patch

even tho we can use --include 'nothing' option and add changes to package.json to the patch.

this wouldn't be a solution, for the reason being it will still run **after install**

which is... well... looses the point, because package hangs at the install stage.

PR #643, opened 2024-01-18
fix: ignore tmp folders in s3 uploading

vsesh:


src/cmd/build/index.ts
```diff
@@ -277,7 +277,7 @@ async function handler(args: Arguments<any>) {
                 await publishFilesToS3({
                     input: userOutputFolder,
                     region: storageRegion,
-                    ignore,
+                    ignore: [...ignore, TMP_INPUT_FOLDER, TMP_OUTPUT_FOLDER],
```
vsesh, line 280: I added tmp folders to be ignored during uploading

yndx-birman (approved):

PR #797, opened 2024-07-05
Implement a configuration option to always specify `vcsPath` for all Md2Md transformations

brotheroftux:
This adds a `vcs` configuration options family, with a single (for now) option available — `remoteBase`. Specifying a correct `vcs.remoteBase` will allow our consumers to have access to Edit button features when using `viewer-external`.

This also opens a possibility to enable the Edit button in standalone builds produced by the `md2html` transformation circuit. This is currently WIP and is not in scope of this PR.


src/services/metadata/enrich.ts
```diff
@@ -0,0 +1,53 @@
...
+type EnrichWithFrontMatterOptions = {
+    fileContent: string;
+    metadataOptions: MetaDataOptions;
+    resolvedFrontMatterVars: FrontMatterVars;
+};
+
+const resolveVCSPath = (metadata: FileMetadata, relativeInputPath: string) => {
+    const maybePreProcessedSourcePath = metadata.sourcePath;
+
+    return typeof maybePreProcessedSourcePath === 'string' && maybePreProcessedSourcePath.length > 0
```
3y3, line 21: Why not?

```
return metadata.sourcePath ? metadata.sourcePath : relativeInputPath;
```

3y3 (approved):

src/services/metadata/enrich.ts
brotheroftux: I'm a bit worried about type safety here — since we're parsing the YAML contents without any additional checks/schema verification procedures, technically `metdata.sourcePath` can be whatever the consumer actually passed as a value, so I erred on the side of caution here. I'm open to discussion on whether this is actually necessary here.

PR #828, opened 2024-09-09
feat: add included option

makamekm:


src/resolvers/md2md.ts
```diff
@@ -83,10 +84,11 @@ function copyFile(targetPath: string, targetDestPath: string, options?: PluginOp
     if (options) {
         const sourceIncludeContent = readFileSync(targetPath, 'utf8');
         const {result} = transformMd2Md(sourceIncludeContent, options);
-
         writeFileSync(targetDestPath, result);
+        return result;
```
martyanovandrey, line 88: wouldn't it be better to read the file separately, not as side effect of the copyFile function ?
for both cases it would be `readFileSync(targetDestPath, 'utf8');`

makamekm: Fixed in https://github.com/diplodoc-platform/transform/pull/509

martyanovandrey (approved):

martyanovandrey (approved):

PR #833, opened 2024-09-17
Re-enable `forceQuotes` while also fixing `systemVars` behaviour

brotheroftux:


Feverqwe (approved):

Feverqwe:
Need `lineWidth: -1` cause liquid variables signature might me broken

PR #896, opened 2024-11-18
fix: check if output parameter is located inside root

martyanovandrey:


src/services/includers/batteries/unarchive.ts
```diff
@@ -87,7 +89,14 @@ async function includerFunction(params: IncluderFunctionParams<Params>) {
 
     const contentPath = index === 0 ? join(writeBasePath, input) : join(readBasePath, input);
 
-    const writePath = join(writeBasePath, output);
+    const writePath = getRealPath(join(writeBasePath, getRealPath(output)));
+
+    if (!writePath.startsWith(writeBasePath)) {
+        throw new UnarchiveIncluderError(
+            `Invalid output parameter: ${output}. Provide includer with output parameter, which is located inside the documentation`,
```
brotheroftux, line 96: I'd say this error message is a bit confusing.
Maybe something along the lines of `Expected the unarchive destination to be located inside project root, got: ${output}`?

brotheroftux (approved):

src/services/includers/batteries/unarchive.ts
main-kun: agree

main-kun (changes requested):
Agree with what @brotheroftux suggested, error message could be better

main-kun (approved):

PR #939, opened 2024-12-11
feat: make collect plugin functions aware of the token stream

brotheroftux:
This integrates https://github.com/diplodoc-platform/transform/pull/581 to utilize the newly implemented `transform.collect` API.

This is a step zero in a plan to make plugin operations isomorphic between their usage in Markdown-to-Markdown and Markdown-to-HTML processing tracks.


main-kun (changes requested):
pls test

main-kun (approved):

Issue #890, opened 2024-11-15
Incorrect quote language in multilingual assembly


xdan:
When you build multilingual documentation ([ru, en]) with the --static-content flag, Russians get into the English documentation Notes

<img width="451" alt="image" src="https://github.com/user-attachments/assets/4e09dedd-cbcd-4822-81ec-ea3a22db886d">

When you build only the English version with the --lang=en flag, there is no such problem

3y3:
The notes plugin supports the `lang` option. You can find the source code for this option here: https://github.com/diplodoc-platform/transform/blob/master/src/transform/plugins/notes/index.ts#L49.

However, somewhere in the process from the command line to the transform to the notes plugin, this option was not passed, and instead, the default value was used.

PR #943, opened 2024-12-14
fix: fixed localisation for notes block

ekiuled:
## Description

Fixed localisation for notes block: pass file language to notes pluging instead of always using yfm args.

Related issue: #890

### DIFF
--- a/src/models.ts
+++ b/src/models.ts
@@ -263,6 +263,7 @@ export interface ResolverOptions {
     fileExtension: string;
     outputPath: string;
     outputBundlePath: string;
+    lang: string;
     metadata?: MetaDataOptions;
 }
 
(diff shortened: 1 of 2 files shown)


martyanovandrey:
Thank you for your contribution! Here are a few points that need to be addressed:

1. I built our documentation with the changes from the current pull request, but I still encountered an error regarding the section titles being in Russian. Could you please let me know which documentation you tested on and with what parameters you ran the build? This information will help us identify the issue more effectively.

2. To facilitate the validation of this PR, it would be helpful to include tests. Tests will ensure that everything is functioning as expected and make it easier for us to proceed with the integration.

3. The `generateStaticMarkup` function is raising an error due to missing type definitions.

Once these issues have been addressed, we can proceed with further review. Thank you for your efforts!

3y3 (approved):

PR #1092, opened 2025-04-02
fix: Fix XSS expluatation in static-content mode

3y3:


src/pages/document.ts
```diff
@@ -65,9 +65,14 @@ export function generateStaticMarkup(
             </head>
             <body class="g-root g-root_theme_light">
                 <div id="root">${html}</div>
+                <script type="application/json" id="state">
```
martyanovandrey, line 68: id="state" may conflict with the title id, i will commit my suggestion in this pr

PR #1120, opened 2025-04-17
fix: add meta, urlTitle, tags, leftItems to translate

goldserg:
1. Add meta to page-constructor
2. Add translate in tags, urlTitle, leftItems.text


goldserg:
merged by PR#1128

PR #1203, opened 2025-05-21
feat: flexible support for restricted-access has been added when adding it to frontmatter

goldserg:
Flexible support for restricted-access has been added when adding it to frontmatter

### DIFF
--- a/src/core/toc/index.spec.ts
+++ b/src/core/toc/index.spec.ts
@@ -264,6 +264,56 @@ describe('toc-loader', () => {
         ),
     );
 
+    it(
+        'should load restricted-access as string',
+        test(
+            dedent`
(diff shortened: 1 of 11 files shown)


src/core/toc/index.spec.ts
```diff
@@ -264,6 +264,56 @@ describe('toc-loader', () => {
...
+                items:
+                  - name: Item without access
+                    href: some/href.md
+            `,
+            {},
+            {},
+        ),
+    );
+    it(
+        'should load restricted-access as array on two level',
```
martyanovandrey, line 298: Could you please add a test to verify how this logic works with nested Tocs?

martyanovandrey (approved):

PR #1271, opened 2025-07-09
feat: supported favicon in static

separatrixxx:
Support for faviconSrc in static mode was added.
Now you can specify in your .yfm file:

```yaml
interface:
  faviconSrc: your-favicon-src
```

The favicon will be displayed correctly, and the icon type will be detected automatically.

### DIFF
--- a/src/core/utils/favicon.ts
+++ b/src/core/utils/favicon.ts
@@ -1,0 +1,19 @@
+export function getFaviconType(faviconSrc: string): string | undefined {
+    if (!faviconSrc) return undefined;
+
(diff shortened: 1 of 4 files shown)


src/core/utils/favicon.ts
```diff
@@ -0,0 +1,19 @@
+export function getFaviconType(faviconSrc: string): string | undefined {
+    if (!faviconSrc) return undefined;
+
+    const ext = faviconSrc.split('.').pop()?.toLowerCase();
```
martyanovandrey: link with query not processed

martyanovandrey (approved):

PR #1284, opened 2025-07-24
feat: added metadata support

separatrixxx:
Added support for metadata on statics. We get the title, description, viewerInterface and resources


src/commands/build/services/entry/EntryService.ts
```diff
@@ -122,6 +135,11 @@ export class EntryService {
             csp.map(template.addCsp);
         }
 
+        if (description && !metadata.some((meta: Hash) => meta.name === 'description')) {
+            metadata.push({name: 'description', content: description});
```
martyanovandrey, line 139: can you give more context here - why exactly are we pushing the description here?

PR #1293, opened 2025-08-01
Extend validation of redirects.yaml

brotheroftux:


src/commands/build/services/redirects/RedirectsService.ts
```diff
@@ -14,6 +14,22 @@ import {getHooks, withHooks} from './hooks';
...
+const EXT_MATCH = /\.(md|yaml)/;
+
+const getRegexpFromRedirect = (match: string) => {
+    let result = match;
+
+    if (!match.startsWith('^')) {
+        result = `^${match}`;
+    }
+
+    if (!match.endsWith('$')) {
```
goldserg, line 26: тут либо одно либо другое? одновременно не может быть?

main-kun (dismissed):
LGTM

src/commands/build/services/redirects/RedirectsService.ts
```diff
@@ -14,6 +14,22 @@ import {getHooks, withHooks} from './hooks';
 
 export const REDIRECTS_FILENAME = 'redirects.yaml';
 
+const EXT_MATCH = /\.(md|yaml)/;
+
+const getRegexpFromRedirect = (match: string) => {
```
main-kun, line 19: This function is problematic. On the first glance it looks like it adds both "$" and "^" as @goldserg noted, but only one of the conditions can be applied because the function only modifies the result once. I think it's confusing and it should either always modify the result if it should apply both, or explicitly apply one or another.

main-kun (approved):
Now LGTM for real

PR #1303, opened 2025-08-06
fix: removed interface and resources from html meta

separatrixxx:
Fixed meta sorting and remove interface and resources from it

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -58,6 +58,6 @@ export enum ResourceType {
     csp = 'csp',
 }
 
-export const testedMetaFields = ['interface', 'resources'];
+export const excludedMetaFields = ['interface', 'resources'];
 
 export const PAGE_PROCESS_CONCURRENCY = Number(process.env.PAGE_PROCESS_CONCURRENCY) || 300;
--- a/src/core/utils/meta.ts
+++ b/src/core/utils/meta.ts
@@ -1,9 +1,9 @@
-import {testedMetaFields} from '~/constants';
+import {excludedMetaFields} from '~/constants';
 
 export function filterMeta<T>(meta: Hash<T>[]): Hash<T>[] {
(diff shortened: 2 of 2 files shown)


src/constants.ts
```diff
@@ -58,4 +58,10 @@ export enum ResourceType {
     csp = 'csp',
 }
 
+export enum YfmFields {
+    interface = 'interface',
+    resources = 'resources',
+    docsViewer = 'docs-viewer',
```
martyanovandrey: In static, we don't use the docs-viewer field, it's only for viewers

let's more clearly name the variable that these are fields that should not be in the html meta tag and leave the rest

ideally we would have a white list of fields that can be here but I'm afraid it will be difficult to support without breaking backward compatibility

martyanovandrey:
and you need to add tests for sorting the required fields

src/constants.ts
separatrixxx: Right now we only support the interface and resources through meta, let's just leave them

src/commands/build/services/entry/EntryService.ts
```diff
@@ -139,7 +139,6 @@ export class EntryService {
             metadata.push({name: 'description', content: description});
         }
 
-        metadata.sort((a: Hash, b: Hash) => a.name.localeCompare(b.name));
```
martyanovandrey, line 142: it was discussed that sorting is needed to pass tests before this after adding the description, what changed?

src/constants.ts
```diff
@@ -58,4 +58,9 @@ export enum ResourceType {
     csp = 'csp',
 }
 
+export enum YfmFields {
```
martyanovandrey: Let's correct the name here; this variable doesn't list all YFM fields—it's more like excludedMetaFields. Also, let's make it an array.

martyanovandrey (approved):

PR #1337, opened 2025-08-18
deps: updated client to 3.9.0

separatrixxx:
fixed e2e tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "5.4.1",
       "license": "MIT",
       "dependencies": {
-        "@diplodoc/client": "^3.8.3",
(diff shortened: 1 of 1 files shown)


tests/e2e/__snapshots__/search.test.ts.snap
```diff
@@ -3,7 +3,6 @@
 exports[`Local search > internal > filelist 1`] = `
 "[
   ".yfm",
-  "_bundle/search-async-1",
```
separatrixxx, line 6: Bundles search-async-0 and search-async-1 are actually scripts 350.js and 969.js in mosks/search/output/_bundle. These are two providers: local and algolia. In client version 3.9.0 we removed the algolia provider, since it completely replicates the local one

PR #1349, opened 2025-08-20
fix: resolve warning punycode

goldserg:


src/require.ts
```diff
@@ -6,4 +6,5 @@ Module.prototype.require = ((original, map) =>
         return original.call(this, map[name] || name);
     }, original))(Module.prototype.require, {
     '@diplodoc/cli': resolve(__dirname, '..'),
+    'punycode': 'punycode/'
```
main-kun: pls fix the lint error

main-kun (changes requested):

PR #1354, opened 2025-08-22
feat: added inline-code plugin

separatrixxx:
added inline-code plugin


martyanovandrey (approved):

martyanovandrey:
Before merging, check that the styles and js of the inline code are connected correctly

PR #1391, opened 2025-09-09
feat: added checkBundle to compareDirectories

separatrixxx:
Added checkBundle to compareDirectories. If this option is set, the list of bundles is not checked


tests/e2e/__snapshots__/metadata.spec.ts.snap
```diff
@@ -1,31 +1,5 @@
...
-  "_bundle/search-async-1",
-  "_bundle/search-css-0",
-  "_bundle/search-css-1",
-  "_bundle/search-css-2",
-  "_bundle/search-css-3",
-  "_bundle/search-css-4",
-  "_bundle/search-js-0",
-  "_bundle/search-js-1",
-  "_bundle/search-js-2",
-  "index.html",
```
martyanovandrey, line 22: we don't look at files in the system `_` folders but we still need to be able to look at other files, let's make that configurable too, default for system files is off, for other files is on

tests/e2e/__snapshots__/load-custom-resources.spec.ts.snap
```diff
@@ -3,25 +3,6 @@
 exports[`Allow load custom resources > md2html single page with custom resources > filelist 1`] = `
 "[
   ".yfm",
-  "_assets/script/test1.js",
```
martyanovandrey, line 6: In this test we check that our custom scripts and styles will load

tests/fixtures/utils/file.ts
```diff
@@ -5,14 +5,38 @@ import {glob} from 'glob';
 import {bundleless, hashless, platformless} from './test';
 import {expect} from 'vitest';
 
+const SYSTEM_DIRS = ['_bundle/', '_assets/', '_search/'];
```
martyanovandrey: _assets should be excluded from system dirs

reazy015 (approved):

PR #1393, opened 2025-09-10
fix: sorted filesForSnapshot

separatrixxx:
Sorted filesForSnapshot

### DIFF
--- a/tests/fixtures/utils/file.ts
+++ b/tests/fixtures/utils/file.ts
@@ -55,6 +55,8 @@ export async function compareDirectories(
         filesForSnapshot = filesFromOutput.filter(file => uselessFile(file, SYSTEM_DIRS));
     }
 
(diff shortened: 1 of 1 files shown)


tests/fixtures/utils/file.ts
```diff
@@ -55,7 +55,9 @@ export async function compareDirectories(
         filesForSnapshot = filesFromOutput.filter(file => uselessFile(file, SYSTEM_DIRS));
     }
 
-    expect(hashless(JSON.stringify(filesForSnapshot, null, 2))).toMatchSnapshot('filelist');
+    filesForSnapshot = filesForSnapshot.map(hashless).sort();
```
martyanovandrey, line 62: do we still need the first sort?

martyanovandrey (approved):

PR #1420, opened 2025-09-16
Deterministic file order (in the TOC) for Generic Includer

khoden:
## Description

Fix random order in the TOC, when using Generic Includer (doc readers may be confused).

Sort from the shortest paths to longest (by path segments).

Put `index.md` as the first item in the TOC (it seems to be useful for the most of use cases).

### DIFF
--- a/src/extensions/generic-includer/index.ts
+++ b/src/extensions/generic-includer/index.ts
@@ -8,7 +8,7 @@ import type {
     YfmString,
 } from '@diplodoc/cli/lib/toc';
 
-import {dirname, extname, join} from 'node:path';
+import {dirname, extname, join, sep} from 'node:path';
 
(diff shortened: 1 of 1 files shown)


src/extensions/generic-includer/index.ts
```diff
@@ -41,10 +41,34 @@ export class Extension implements IExtension {
                         cwd: join(run.input, input),
                     });
 
-                    return fillToc(toc, graph(files), options);
+                    return fillToc(toc, graph(this.sortGlobResult(files)), options);
                 });
         });
     }
+
+    sortGlobResult(files: NormalizedPath[]): NormalizedPath[] {
```
martyanovandrey, line 49: please add tests for this behavior

PR #1427, opened 2025-09-18
fix: support reference links in autotitles

goldserg:
fix: support reference links in autotitles
fix: incorrect rebased

### DIFF
--- a/src/commands/build/features/output-md/plugins/merge-autotitles.ts
+++ b/src/commands/build/features/output-md/plugins/merge-autotitles.ts
@@ -3,9 +3,13 @@ import type {Run} from '../../..';
 import type {StepFunction} from '../utils';
 
 import {dirname, join, relative} from 'node:path';
+import {normalizePath} from '~/core/utils';
 
 function isAutotitle(asset: AssetInfo) {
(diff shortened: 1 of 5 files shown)


tests/mocks/preprocess/input/autotitle.md
reazy015, line 1: There is possible case when autotile goes afterward inner anchor it refers to

src/commands/build/features/output-md/plugins/merge-autotitles.ts
```diff
@@ -16,13 +20,15 @@ function rebaseAssets(
 ) {
     const base = relative(dirname(to), dirname(from));
     return assets.reduceRight((content, {path, location}) => {
-        const link = join(base, path);
-        return content.slice(0, location[0]) + link + content.slice(location[1]);
+        const link = normalizePath(join(base, path));
```
reazy015, line 23: here in case if path is null we can catch TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received null

This happens when autotitle goes afterward inner anchor it refers to in the same md file

PR #1430, opened 2025-09-20
feat: added pages to state

separatrixxx:
.yfm supports the pages field, which currently accepts a setting for the accessDenied page: the text field can change the default text on the 403 page for different locales


separatrixxx:
This is only relevant for viewers; there is no need to support static content.

PR #1431, opened 2025-09-22
fix: rollback resolve codeblock

goldserg:


src/core/markdown/loader/resolve-assets.ts
```diff
@@ -11,7 +11,7 @@ export function resolveAssets(this: LoaderContext, content: string) {
     const exclude = [
         ...this.api.deps.get().map(({location}) => location),
         ...this.api.comments.get(),
-        ...this.api.blockCodes.get(),
+        // ...this.api.blockCodes.get(),
```
martyanovandrey: Let's do it through a regular revert instead of commenting the code.

src/core/markdown/__snapshots__/loader.spec.ts.snap
```diff
@@ -121,13 +121,41 @@ exports[`Markdown loader > resolveAssets > should detect link to anchor 1`] = `
 
 exports[`Markdown loader > resolveAssets > should not detect asset with language and translate=no directive 1`] = `
 [
-  [],
+  [
```
martyanovandrey, line 124: Why were there no test changes in the version where the line was added, but here there are?

https://github.com/diplodoc-platform/cli/commit/82f93619ac59bdf1e2a11b5cc626e4df69a9ed2e#diff-4814f8a7633d153c51afedc0acf3c5044341a890d9ef038404c3be980efd4fa8R11-R14

PR #1447, opened 2025-09-24
fix: Added a pdf-page feature to assist with pdf generation

separatrixxx:
Added a pdf-page feature to assist with pdf generation. Now, despite the singlepage, separate system files are created for generating pdf.

### DIFF
--- a/src/commands/build/features/pdf-page/index.ts
+++ b/src/commands/build/features/pdf-page/index.ts
@@ -30,12 +30,11 @@ export class PdfPage {
         getBuildHooks(program)
             .Entry.for('html')
             .tap('PdfPage', (run, entry, info) => {
-                if (!run.config.preparePdf || !info.html) {
+                if (!run.config?.pdf?.enabled || !info.html) {
                     return;
                 }
 
-                const hiddenPolicy = run.config.hiddenPolicy;
-                const isHiddenPolicy = hiddenPolicy?.pdf ?? true;
+                const isHiddenPolicy = run.config?.pdf?.hiddenPolicy ?? true;
 
                 const toc = run.toc.for(entry);
                 const meta = info.meta || {};
@@ -59,7 +58,7 @@ export class PdfPage {
         getBuildHooks(program)
             .BeforeRun.for('html')
             .tap('PdfPage', (run) => {
-                if (!run.config.preparePdf) {
+                if (!run.config?.pdf?.enabled) {
                     return;
                 }
 
@@ -91,7 +90,7 @@ export class PdfPage {
         getBuildHooks(program)
             .AfterRun.for('html')
             .tapPromise('PdfPage', async (run) => {
-                if (!run.config.preparePdf) {
+                if (!run.config?.pdf?.enabled) {
                     return;
                 }
 
--- a/src/commands/build/types.ts
+++ b/src/commands/build/types.ts
@@ -46,9 +46,8 @@ type BaseConfig = {
     // TODO: explicitly handle
     analytics: DocAnalytics;
     supportGithubAnchors?: boolean;
-    interface: Hash;
-    hiddenPolicy: Record<string, boolean>;
-    preparePdf: boolean;
+    interface?: Hash;
+    pdf?: Record<string, boolean>;
 };
 
 export type VcsArgs = {


src/commands/build/features/pdf-page/index.ts
```diff
@@ -0,0 +1,150 @@
...
+                    return;
+                }
+
+                for (const toc of run.toc.tocs) {
+                    const entries: PdfPageResult[] = [];
+
+                    const hiddenPolicy = run.config.hiddenPolicy;
+                    const isHiddenPolicy = hiddenPolicy?.pdf ?? true;
+
+                    if (isHiddenPolicy) {
```
reazy015: Is it possible to simplify this condition by moving isHiddenPolicy inside walkItems iteration?

For example

if (isHiddenPolicy  && (item.hidden || !item.href)) return;

src/commands/build/features/pdf-page/index.ts
```diff
@@ -0,0 +1,150 @@
...
+const __PdfPage__ = Symbol('isPdfPage');
+
+export class PdfPage {
+    apply(program: Build) {
+        const results: Record<NormalizedPath, PdfPageResult> = {};
+
+        getBuildHooks(program)
+            .Entry.for('html')
+            .tap('PdfPage', (run, entry, info) => {
+                if (!run.config['docs-viewer'].pdf || !info.html) {
```
reazy015: Not critical btw.

Maybe it's better to create separate prop out of docs-viewer scope. This one we're using for making decision of pdf icon rendering.

src/commands/build/types.ts
```diff
@@ -47,6 +47,10 @@ type BaseConfig = {
     analytics: DocAnalytics;
     supportGithubAnchors?: boolean;
     interface: Hash;
+    hiddenPolicy: Record<string, boolean>;
+    'docs-viewer': {
```
martyanovandrey: `docs-viewer` should not be used in cli, only in viewer

tests/e2e/__snapshots__/pdf-page.spec.ts.snap
```diff
@@ -0,0 +1,172 @@
...
+
+exports[`Pdf page mode > transforms links correctly > filelist 1`] = `
+"[
+  ".yfm",
+  "index.html",
+  "page1.html",
+  "page2.html",
+  "page3.html",
+  "pdf-page-toc.js",
+  "pdf-page.html",
```
martyanovandrey: If, when generating a single page, the files were the final result, now they're intermediate system files.

1) They should either be placed in a separate system folder or start with _ - see what works best.
2) There should definitely be an option to not create them at all or to delete them after creation once we learn how to generate PDFs using an extension. We must make it clear to users that these files are not for use, but are needed only for creating PDFs.

src/commands/build/features/pdf-page/utils.ts
```diff
@@ -0,0 +1,18 @@
+import type {SinglePageResult} from '../singlepage/utils';
+
+import {getSinglePageUrl, joinSinglePageResults} from '../singlepage/utils';
```
martyanovandrey, line 7: We discussed how it's best not to make one feature dependent on another. I'm suggesting creating a page meta feature and then two subfeatures within it: a single page and a PDF.

If you have any other ideas for how to make this feature work, please discuss them. Current dependencies will be difficult to maintain later.

// features/pages/base.ts

```ts
export interface PageFeatureOptions {
  targetFile: string;
  entryType: 'single-page' | 'pdf-page';
}

export class PageFeatureCore {
  constructor(private opts: PageFeatureOptions) {}

}
```

```ts
export class SinglePage extends PageFeatureCore {
  constructor() {
    super({
      targetFile: 'single-page.html',
      entryType: 'single-page',
      ...
    });
  }
}

export class PdfPage extends PageFeatureCore {
  constructor() {
    super({
      targetFile: 'pdf-page.html',
      entryType: 'pdf-page',
      ...
    });
  }
}
```

martyanovandrey (dismissed):

src/commands/build/features/pdf-page/utils.ts
separatrixxx: On the one hand, pdf-page currently conceptually copies and uses some of singlepage's utilities. It seems wrong for singlepage to inherit a basic feature that won't be used anywhere else. It also seems wrong for pdf-page to inherit from singlepage if we plan to expand the feature in the future.

Currently, pdf-page only uses two utilities from singlepage, one of which will rewrite (mentioned in the comments). Creating a new feature just to import one function, which might also need to be rewritten, also seems wrong.

reazy015 (approved):

PR #1448, opened 2025-09-25
fix: build manifest

goldserg:


src/commands/build/features/build-manifest/index.ts
```diff
@@ -150,13 +150,18 @@ export class BuildManifest {
 
             if (lastHead[name]?.file) {
                 const pathToReport = path.replace(/\..+$/, '');
+                const existingExt = lastHead[name]?.file?.ext;
+                const shouldReplace = this.shouldReplaceFile(existingExt, ext);
```
martyanovandrey, line 159: Can you please add tests for this functionality? and add a description in the PR of what and why we are changing

goldserg: done

src/commands/build/features/build-manifest/index.ts
```diff
@@ -177,6 +182,20 @@ export class BuildManifest {
         };
     }
 
+    private shouldReplaceFile(existingExt: string, newExt: string): boolean {
```
martyanovandrey, line 190: So, how will we behave if we visit an existing index.yaml page but also have an index.md file? It turns out we'll always be redirected to index.md in viewers?

PR #1455, opened 2025-10-03
fix: added arg --pdf

separatrixxx:
In addition to the .yfm pdf option: enabled: true , the --pdf arg is supported


src/commands/build/features/pdf-page/index.ts
```diff
@@ -9,9 +10,17 @@ import {getHooks as getBaseHooks} from '~/core/program';
...
-import {defined} from '~/core/config';
 
 import {PDF_PAGE_FILENAME, getPdfUrl, isEntryHidden, joinPdfPageResults} from './utils';
+import {options} from './config';
+
+export type PdfPageArgs = {
+    pdf: boolean;
+};
+
+export type PdfPageConfig = {
```
reazy015: Looks strange and inconsistently in my opinion

PdfPageConfig type with singlePage record, personally I would be confused). 

And btw i don't see any place where we use it

PR #1846, opened 2026-04-15
feat(analytics): add csp rules when metrika counter is configured

stenin-nikita:


src/commands/build/features/analytics/constants.ts
```diff
@@ -0,0 +1,10 @@
+const METRIKA_URL = 'https://mc.yandex.ru';
+
+export const METRIKA_CSP_RULES = {
+    'script-src': [METRIKA_URL],
```
martyanovandrey, line 4: will these csp be enough?

martyanovandrey (approved):

src/commands/build/features/analytics/constants.ts
stenin-nikita: According to the documentation https://yandex.ru/support/metrica/ru/code/install-counter-csp this should be sufficient

PR #1850, opened 2026-04-16
fix: copy pdf icon as custom resource

reazy015:


src/commands/build/features/custom-resources/copy-docs-viewer-pdf-icon-asset.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+        warn: vi.fn(),
+        error: vi.fn(),
+    };
+
+    return {
+        input: '/project/input',
+        output: '/project/output',
+        config: {
+            content: {maxAssetSize},
+            ...(docsViewer === undefined ? {} : {'docs-viewer': docsViewer}),
```
martyanovandrey: the `docs-viewer` block is only for viewers, we don't use it in open source, and the feature should work for static data

src/commands/build/features/custom-resources/docs-viewer-pdf-icon-path.ts
```diff
@@ -0,0 +1,48 @@
...
+        return;
+    }
+
+    const icon = (pdf as {icon?: unknown}).icon;
+    if (typeof icon !== 'string') {
+        return;
+    }
+
+    const trimmed = icon.trim();
+    if (!trimmed.startsWith('_assets')) {
```
martyanovandrey: _assets-backup/leaked.png
_assetsX/icon.svg

goldserg (approved):

src/commands/build/features/custom-resources/index.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+            typeof run.fs.statSync
+        >);
+
+        await getHooks(build).AfterRun.for('md').promise(run);
+
+        expect(run.copy).toHaveBeenCalledWith(assetPath, join(run.output, '_assets/icons/pdf.svg'));
+        expect(run.copy).toHaveBeenCalledWith(stylePath, join(run.output, '_assets/style/a.css'));
+    });
+
+    it('AfterRun md: copies pdf icon even when allowCustomResources is off', async () => {
```
martyanovandrey: Are we sure this is the right place this feature implementation?

The logic for copying the pdf icon seems to live in the wrong feature - custom-resources. copyPdfIconAsset is triggered even when allowCustomResources: false (and this is covered by a test), so the customResources feature is doing work that is not actually related to custom resources.

Why isn't it a completely separate part of features/pdf-page/? Otherwise, if CustomResources is disabled or refactored later, it would be easy to accidentally break PDF icon copying.

martyanovandrey (approved):

PR #1883, opened 2026-04-27
fix: Support includes in terms

goldserg:


src/commands/build/features/output-html/index.ts
```diff
@@ -60,6 +61,11 @@ export class OutputHtml {
                     });
                 });
 
+                getMetaHooks(run.meta).Dump.tap('Html', (meta) => {
+                    filterBundledExtensionAssets(meta);
+                    return meta;
```
stenin-nikita: The function mutates the meta object internally. In most cases, it's preferable to use pure functions — is mutation actually necessary here?

adr/ADR-006-merge-includes-md2md.md
```diff
@@ -1215,26 +1231,254 @@ Source map комментарии добавляются в `prepareInlinedConte
...
 
 ### Risks
 
-⚠️ Конфликты якорей и ID
-⚠️ Неожиданное поведение при сложных вложенностях
+⚠️ Конфликты якорей и ID (warning-и при обнаружении, полная детекция — Этап 7)
+⚠️ Неожиданное поведение при сложных вложенностях (протестировано на реальных доках)
 ⚠️ Производительность при глубокой вложенности includes
 
+## Known Bugs (найдены и исправлены)
```
martyanovandrey: it would be better to use english in opensource

martyanovandrey (approved):

PR #1888, opened 2026-04-27
fix: skiped error for arcadia ci

Yurok868:
fix: skiped error for arcadia ci


src/core/program/index.ts
```diff
@@ -223,6 +223,14 @@ export class BaseProgram<
                 return;
             }
 
+            const message = errorMessage(error);
```
martyanovandrey: the error should be caught not globally, but at the initialization point
https://github.com/diplodoc-platform/cli/blob/0f36c8d2e44923c145ad6c33460799cd38080c08/src/extensions/arcadia-vcs/connector.ts#L32

something like

```
    try {
        await Promise.all(
            [
                mtimes.enabled && this.fillMTimes(),
                authors.enabled && this.fillAuthors(),
                contributors.enabled && this.fillContributors(),
            ].filter(Boolean),
        );
    } catch (error) {
        if (
            error &&
            typeof error === 'object' &&
            'code' in error &&
            (error as NodeJS.ErrnoException).code === 'ENOENT'
        ) {
            // eslint-disable-next-line no-console
            console.warn(
                'Arcadia VCS extension disabled: arc is not available in this environment.',
            );
            return this;
        }
        throw error;
    }
```

and we need to add tests to this place.

martyanovandrey (changes requested):

PR #1894, opened 2026-04-28
fix(template): preserve whitespace in body interpolation in Template.dump

### LINKED ISSUE #1893: Bug: extra 8-space indentation in `<pre><code>` blocks since `@diplodoc/cli@5.0.0-rc-8`
### Package version

5.36.2

### Node version

v24.15.0

### Platform

Darwin arm64 (uname -mo)

## Problem

Starting from `@diplodoc/cli@5.0.0-rc-8`, every line **after the first one**
inside `<pre><code>` blocks in the generated HTML gets an extra **8 spaces** of
indentation. The first line (the one that ends up on the same line as the
opening `<code>` tag) is unaffected.

The original markdown source has no leading whitespace on those lines.

## Regression range

| Version                 | Result             |
| ----------------------- | ------------------ |
| `4.57.22`               | ✅ no bug           |
| `5.0.0-rc-1` … `rc-4`   | ✅ no bug           |
| **`5.0.0-rc-8`**        | ❌ **first broken** |
| `5.0.0-rc-9` … `5.36.0` | ❌ broken           |

(`4.57.23–4.60.0` and `5.0.0-rc-7` could not be tested due to unrelated
`staticContent` config incompatibilities.)

## Root cause

The `Template.dump()` method introduced in rc-8
([`@diplodoc/cli/lib/template/index.js`](node_modules/@diplodoc/cli/lib/template/index.js:1),
source `src/core/template/template.ts`) builds the final HTML page using a
`ts-dedent` template literal:

```js
return dedent`
            <!DOCTYPE html>
            <html lang="${lang}" ...>
                ...
                <body class="${bodyClass.join(" ")}">
                    ${body.join("\n") || `<div id="root"></div>`}
                    ...
                </body>
            </html>
        `;
```

`body` contains the SSR-rendered `<div id="root">${html}</div>`, which is a
multi-line HTML string. `ts-dedent` strips the common leading indent from the
**static** parts of the literal (8 spaces here). When a multi-line interpolation
is inserted, the lines after the first inherit the literal column position of
`${...}` (20 spaces) minus the common indent (8 spaces) = **8 spaces** prepended
to every subsequent line — which destroys whitespace-significant content like
`<pre><code>`.

### Minimal reproduction

```js
const { dedent } = require('ts-dedent');
const body = '<pre><code>line1\nline2\nline3</code></pre>';
console.log(dedent`
            <body>
                ${body}
            </body>
        `);
```

Output:

```
<body>
    <pre><code>line1
        line2
        line3</code></pre>
</body>
```

This matches the bug exactly.

### Evidence from real build output

- `state.data.html` (input to SSR, extracted from `<script id="diplodoc-state">`):
  `./init.sh --dump-postgres` — **0 leading spaces** ✅
- Rendered `<div id="root">…</div>` (output of `Template.dump()`):
  `        ./init.sh --dump-postgres` — **8 leading spaces** ❌

So all stages up to `Template.dump()` are clean; the indentation is added when
the page wrapper is assembled.

## Suggested fix

Stop using `dedent` (or any common-indent-stripping tagged template) to splice
multi-line, whitespace-significant content. Options:

1. Build the outer page wrapper with `dedent`, but concatenate `body` outside of
   it (so its newlines aren't reflowed by the dedent algorithm).
2. Drop `ts-dedent` for `Template.dump()` and write the wrapper plainly.
3. Pre-process `body` to insert a prefix that compensates for the dedent
   behavior — fragile, not recommended.

### Expected output (`4.57.22` and earlier)

```html
<div class="yfm-clipboard">
    <pre><code class="hljs bash"><span class="hljs-comment"># 1. Создать резервную копию</span>
./init.sh --dump-postgres

<span class="hljs-comment"># 2. Подключиться к PostgreSQL</span>
./init.sh --pg-sql
</code></pre>
</div>
```

### Actual output (`5.0.0-rc-8` … `5.36.0`)

```html
<div class="yfm-clipboard">
    <pre><code class="hljs bash"><span class="hljs-comment"># 1. Создать резервную копию</span>
        ./init.sh --dump-postgres
        
        <span class="hljs-comment"># 2. Подключиться к PostgreSQL</span>
        ./init.sh --pg-sql
        </code></pre>
</div>
```

### Additional information

_No response_

goshander:
## Summary

Fixes #1893.

`Template.dump()` (introduced in `5.0.0-rc-8`) builds the final page using a `ts-dedent` template literal. `ts-dedent` strips the common leading indent from the literal but **also re-indents every line of a multi-line interpolation** to the column of the `${...}` placeholder.

`body.join('\n')` is interpolated 8 spaces deep inside `<body>`, so every line **after the first** of the SSR-rendered body markup gets 8 extra spaces prepended. For most tags this is invisible, but inside whitespace-significant content like `<pre><code>` blocks the extra indent shows up verbatim in the rendered page:

```html
<!-- before -->
<pre><code class="hljs bash">./first-line.sh
        ./second-line.sh        <!-- 8 extra spaces -->
        ./third-line.sh         <!-- 8 extra spaces -->
</code></pre>
```

## Fix

Insert `body.join('\n')` via a placeholder (`\x00BODY\x00`) and substitute it with `.replace()` **after** `dedent` has finished processing, so the body's newlines stay untouched. Body is the only whitespace-significant interpolation in the template, so other interpolations don't need protection.

## Reproduction

Standalone `ts-dedent` repro (10 lines), regression range `5.0.0-rc-4` (clean) → `5.0.0-rc-8` (broken), and full evidence are in the linked issue (#1893).

## Verification

Built `lib/template` from the patched source, dropped it into `node_modules/@diplodoc/cli/lib/template/` of a real docs project, and ran a full `npx @diplodoc/cli` build. Generated `<pre><code>` blocks now contain content with 0 leading spaces — matching the `4.x` behaviour.

Closes #1893

### DIFF
--- a/src/core/template/template.ts
+++ b/src/core/template/template.ts
@@ -339,6 +339,11 @@ export class Template {
         const base = getDepthPath(getDepth(this.path) - 1);
         const faviconType = getFaviconType(faviconSrc);
 
+        // ts-dedent re-indents every line of a multi-line interpolation,
+        // which corrupts whitespace-significant content (e.g. <pre><code> blocks) inside body
+        // insert body via a placeholder so its newlines stay untouched
+        const BODY = '\x00BODY\x00';
+
         return dedent`
             <!DOCTYPE html>
             <html lang="${lang}" dir="${this.isRTL ? 'rtl' : 'ltr'}">
@@ -357,13 +362,13 @@ export class Template {
                     ${leading(styles).map(style(this.csp)).join('\n')}
                 </head>
                 <body class="${bodyClass.join(' ')}">
-                    ${body.join('\n') || `<div id="root"></div>`}
+                    ${BODY}
                     ${state(scripts).map(script(this.csp)).join('\n')}
                     ${trailing(scripts).map(script(this.csp)).join('\n')}
                     ${trailing(styles).map(style(this.csp)).join('\n')}
                 </body>
             </html>
-        `;
+        `.replace(BODY, body.join('\n') || `<div id="root"></div>`);
     }
 }
 


src/core/template/template.ts
```diff
@@ -357,13 +362,13 @@ export class Template {
...
                 <body class="${bodyClass.join(' ')}">
-                    ${body.join('\n') || `<div id="root"></div>`}
+                    ${BODY}
                     ${state(scripts).map(script(this.csp)).join('\n')}
                     ${trailing(scripts).map(script(this.csp)).join('\n')}
                     ${trailing(styles).map(style(this.csp)).join('\n')}
                 </body>
             </html>
-        `;
+        `.replace(BODY, body.join('\n') || `<div id="root"></div>`);
```
main-kun, line 371: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_the_replacement

В теле вполне могут быть $$

Лучше передать функцию как замену

main-kun (changes requested):

goshander:
@main-kun перенес в PR без форка
https://github.com/diplodoc-platform/cli/pull/1918

PR #1898, opened 2026-04-29
feat(generic-includer): add linkIndexAutotitle option for directory titles from index.md

### LINKED ISSUE #1872: Отображать заголовок родительского раздела из index.md в GenericIncluder
### What is the problem this feature will solve?

Автоматически синхронизировать название заголовков родительских страниц/директорий с их названием в структуре. На данный момент такой возможности нет.

### What is the feature you are proposing to solve the problem?

При включенной опции [linkIndex ](https://github.com/diplodoc-platform/cli/commit/bb3ee3908ef5042ead8f7ac01771808da170a8fc) и autotitle в GenericIncluder по аналогии с обычными страницами для директорий отображать название директории в структуре из заголовка в вложенном в директорию `index.md`

Image

Image 2

### What alternatives have you considered?

Использовать физическое название файлов для наименование структуры разделов.
Но этот вариант порождает ограничения и неудобства:
- Нельзя использовать в название файла символы вроде точки
- Потребуется поддерживать в идентичном виде название файлов для структуры и заголовок для самой страницы

martyanovandrey:
## Summary
- Add `linkIndexAutotitle` option (default `false`) to the `generic` includer.
- When `linkIndexAutotitle: true` is set together with `linkIndex` (and `autotitle` is not disabled), the directory entry's title is derived from the H1 heading of its nested `index.md` via the existing autotitle pipeline. Otherwise the folder name is used (backward-compatible default).

Closes #1872


src/extensions/generic-includer/index.ts
```diff
@@ -90,17 +91,31 @@ function fillToc(toc: RawToc, graph: Graph, options: Options) {
             const entries = Object.entries(value);
             const indexEntry = entries.find(([k]) => k === 'index');
             const childEntries = entries.filter(([k]) => k !== 'index');
+            const indexHref =
+                indexEntry && typeof indexEntry[1] === 'string' ? indexEntry[1] : undefined;
+
+            if (indexHref) {
+                const useIndexHeading =
+                    options.linkIndexAutotitle === true && options.autotitle !== false;
```
separatrixxx, line 129: Why not options.linkIndexAutotitle && !options.autotitle?

src/extensions/generic-includer/index.ts
```diff
@@ -90,17 +91,31 @@ function fillToc(toc: RawToc, graph: Graph, options: Options) {
...
+
+                if (useIndexHeading) {
```
separatrixxx, line 135: const res = {
    href: indexHref,
    items: childEntries.map(item),
}

if (useIndexHeading) {
    return res;
}

return {
    name: key as YfmString,
    ...res,
}

mb tak?

src/extensions/generic-includer/index.ts
martyanovandrey: The suggestion inverts the semantics. `options.autotitle !== false` is true when autotitle is enabled (`true` or `undefined`), while `!options.autotitle` is true when it's disabled (`false` or `undefined`). We need the former — deriving a title from `index.md`'s H1 only makes sense when autotitle isn't disabled. The `should keep folder name when linkIndexAutotitle is set but autotitle is disabled` test covers this.

separatrixxx (approved):

PR #1905, opened 2026-04-30
feat(build): emit build stats file with timing and counters

martyanovandrey:
## Summary
- Adds an opt-in `BuildStats` feature that writes `yfm-build-stats.json` next to other build artifacts when `--build-stats` is set (or `buildStats: true` in `.yfm`). Intended for CI dashboards and regression detection, not for the runtime to consume.
- Format-agnostic: hooks into `BeforeAnyRun` / `AfterAnyRun` and the build-scoped `Entry` hook for both `md` and `html`. When the flag is off the feature is fully inert.
- `build.configHash` is sha256 of a stably-serialized `run.config` and matches the cache-key input shape called out in the FileProducer RFC, so it can be reused once that lands.

## What gets recorded
- **`cli`** — `version`, `node`, `platform`, `arch`, `osRelease`
- **`build`** — `startedAt`, `finishedAt`, `durationMs`, and a coarse `phasesMs.{prepare, entries, finalize}` derived from `Entry`-hook timestamps (no new hooks); plus `outputFormat`, `langs`, `inputDir`, `outputDir`, `configHash`, `features` (every config field that resolved to literal `true`), `worker.maxOldSpace`
- **`counters`** — `tocs`, `entriesPlanned` vs `entriesProcessed`, `entriesByExtension`, `entriesByLang`, `missed` (walks `entry.relations` for nodes tagged `type: 'missed'`), `warnings`, `errors`
- **`output`** — `files`, `totalBytes`, `bytesByExtension`, `largestFile` via globbing the output dir; failures to glob/stat log a warning and don't fail the build

`schemaVersion: 1` is at the top level so the format can grow without breaking consumers.

## Out of scope
- Per-entry duration distribution (p50/p95/top-N slowest pages) — needs a hook signature change since the current `Entry` hook fires before the per-entry timing in `Build.processEntry` is computed.
- Precise per-phase timing (`prepareInput`/`prepareRun`/`sync`/`processToc`/`releaseRun`/`cleanup`) — phases are currently just `console.log` calls in `Build.action()`; needs either phase hooks or a `run.timer.mark()` API.

### DIFF
--- a/src/commands/build/features/build-stats/index.ts
+++ b/src/commands/build/features/build-stats/index.ts
@@ -2,7 +2,7 @@ import type {Command} from '~/core/config';
 import type {Build, EntryInfo, Run} from '~/commands/build';
 
 import {extname, join} from 'node:path';
-import {arch, platform, release as osRelease} from 'node:os';
+import {arch, release as osRelease, platform} from 'node:os';
 import {createHash} from 'node:crypto';
 
 import {getHooks as getBaseHooks} from '~/core/program';
--- a/src/commands/build/features/build-stats/index.ts
+++ b/src/commands/build/features/build-stats/index.ts
@@ -214,7 +214,7 @@ export class BuildStats {
     }
 }
 
-export function toLangCode(lang: string | {lang: string}): string {
+function toLangCode(lang: string | {lang: string}): string {
     return typeof lang === 'string' ? lang : lang.lang;
 }
 
@@ -237,7 +237,7 @@ function countMissed(run: Run): number {
     return count;
 }
 
-export function stableStringify(value: unknown): string {
+function stableStringify(value: unknown): string {
     if (value === null || typeof value !== 'object') {
         return JSON.stringify(value) ?? 'null';
     }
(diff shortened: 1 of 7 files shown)


src/commands/build/features/build-stats/index.ts
```diff
@@ -0,0 +1,363 @@
...
+            const ext = (extname(entry) || '<noext>').toLowerCase();
+            this.entriesByExtension[ext] = (this.entriesByExtension[ext] ?? 0) + 1;
+
+            const lang = langFromPath(entry, run.config) || '<none>';
+            this.entriesByLang[lang] = (this.entriesByLang[lang] ?? 0) + 1;
+
+            // Markdown-only payload: leading (yaml) pages don't have headings/html.
+            if (info.leading === false) {
+                this.headings += info.headings?.length ?? 0;
+                this.contentBytes += info.html?.length ?? 0;
```
main-kun: The length of the string is not bytes since it's unicode.

src/commands/build/features/build-stats/index.ts
```diff
@@ -0,0 +1,363 @@
...
+        files = await run.glob('**/*', {cwd: run.output});
+    } catch (error) {
+        run.logger.warn(`BuildStats: failed to glob output directory: ${error}`);
+        return EMPTY_OUTPUT;
+    }
+
+    let totalBytes = 0;
+    const bytesByExtension: CountByKey = {};
+
+    await Promise.all(
```
main-kun: Need a bound here for concurrency

main-kun (changes requested):

main-kun (approved):

PR #1907, opened 2026-04-30
fix: fixed arcadia-vcs

Yurok868:
fix: fixed arcadia-vcs

### DIFF
--- a/src/extensions/arcadia-vcs/connector.ts
+++ b/src/extensions/arcadia-vcs/connector.ts
@@ -23,6 +23,8 @@ export class ArcadiaVcsConnector implements VcsConnector {
 
     private run: Run<Config>;
 
+    private arcAvailable = false;
+
     constructor(run: Run<Config>) {
         this.run = run;
         this.config = run.config;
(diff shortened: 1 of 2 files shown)


src/extensions/arcadia-vcs/connector.ts
```diff
@@ -67,7 +70,11 @@ export class ArcadiaVcsConnector implements VcsConnector {
         this.contributorsByPath = data.contributors;
     }
 
-    getBase() {
+    async getBase() {
+        if (!this.arcAvailable) {
```
martyanovandrey, line 74: need unit test that mocks ArcClient to throw {code: 'ENOENT'} from a fill* method, then asserts init() resolves and getBase() returns '.',

src/extensions/arcadia-vcs/connector.ts
```diff
@@ -45,6 +47,7 @@ export class ArcadiaVcsConnector implements VcsConnector {
                 this.run.logger.warn(
                     'Arcadia VCS extension disabled: arc is not available in this environment.',
                 );
+                this.arcAvailable = false;
```
martyanovandrey: the default flag is true, so calling getBase() before init() completes will still go to arc.getBase() and fail. This is safe now because of the registration through BeforeAnyRun, but if getBase() starts being called earlier in the lifecycle, the guard will stop working. It's safer to start with false and switch to true after init() succeeds

martyanovandrey (approved):

PR #1930, opened 2026-05-14
fix: fixed getAuthors

Yurok868:
fix(arcadia-vcs): credit authors on cross-scope commits in getAuthors

### DIFF
--- a/src/extensions/arcadia-vcs/arc-client.ts
+++ b/src/extensions/arcadia-vcs/arc-client.ts
@@ -63,7 +63,6 @@ export class ArcClient {
     async getAuthors(): Promise<Record<string, {login: string; commit: string}>> {
         const base = await this.getBase();
         const scopes = [base, ...(this.config.vcs.scopes || [])].map(normalizePath);
-        const handled = new Set<string>();
 
         const authors: Record<string, {login: string; commit: string}> = {};
 
@@ -73,9 +72,7 @@ export class ArcClient {
             const ignore = this.config.authors?.ignore || [];
 
             for (const commit of commits) {
-                const skip =
-                    shouldBeIgnored(ignore, {login: commit.login}) || handled.has(commit.sha);
-                handled.add(commit.sha);
+                const skip = shouldBeIgnored(ignore, {login: commit.login});
 
                 followPaths(
                     commit.paths,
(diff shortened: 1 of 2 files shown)


src/extensions/arcadia-vcs/arc-client.ts
```diff
@@ -73,9 +72,7 @@ export class ArcClient {
             const ignore = this.config.authors?.ignore || [];
 
             for (const commit of commits) {
-                const skip =
-                    shouldBeIgnored(ignore, {login: commit.login}) || handled.has(commit.sha);
-                handled.add(commit.sha);
+                const skip = shouldBeIgnored(ignore, {login: commit.login});
```
martyanovandrey, line 74: We need tests for this change.

martyanovandrey, line 74: A similar bug will be present in getContributors, where you can use Set<sha+path> instead of Set<sha>.

martyanovandrey (changes requested):

src/extensions/arcadia-vcs/arc-client.ts
```diff
@@ -63,7 +61,6 @@ export class ArcClient {
     async getAuthors(): Promise<Record<string, {login: string; commit: string}>> {
         const base = await this.getBase();
         const scopes = [base, ...(this.config.vcs.scopes || [])].map(normalizePath);
-        const handled = new Set<string>();
```
martyanovandrey, line 66: if you remove this behavior, several cases will break. Add tests for them

1) getAuthors: one commit affects files in different scopes — we check that the author is set for paths from all scopes, not just the first one. This is the main bug that the PR is fixing. It is already in the PR.

2) getContributors: one commit affects files in different scopes — similar to point 1, but for an accumulating array of contributors. It is already in the PR, but the mock for . is unrealistic — we need arc log . returned all commit paths, including those that would be included in the sub-scope.

3) getContributors: one path in the overlapping scopes (. + docs) - arc log . and arc log docs both return a commit with the path docs/intro.md; we check that there is only one entry in the contributors array, not two. This is a regression caused by removing the handled Set from getContributors, and it is directly related to the changes in the PR.

src/extensions/arcadia-vcs/arc-client.ts
```diff
@@ -73,9 +70,7 @@ export class ArcClient {
             const ignore = this.config.authors?.ignore || [];
 
             for (const commit of commits) {
-                const skip =
-                    shouldBeIgnored(ignore, {login: commit.login}) || handled.has(commit.sha);
-                handled.add(commit.sha);
+                const skip = shouldBeIgnored(ignore, {login: commit.login});
```
martyanovandrey, line 75: fix: fixed getAuthors - it doesn't say anything, so I suggest something like fix(arcadia-vcs): credit authors on cross-scope commits in getAuthor

martyanovandrey (approved):

PR #1948, opened 2026-05-22
fix: bugs mergeIncludes

goldserg:
Bug 19: Blank lines inside multi-line HTML comments break list/cut/blockquote when indented
Bug 20: `{% include %}` inside fenced code blocks (in lists) was expanded as a real include
Bug 21: Fence range end collided with the `{% include %}` on the very next line (regression of Bug 20)
Bug 22: Unterminated fence opener swallowed every `{% include %}` to EOF
Bug 23: stray `|` rendered in YFM shorthand cell when an inlined include ends with an HTML block
Bug 24: `|` inside inlined content split YFM shorthand cells
Bug 25: ` ``` |` (fence + YFM cell separator on the same line) was rejected as a closer
Bug 26: fence runs inside an HTML comment paired across the comment boundary
Bug 28: end-of-line ` ```|| ` inside a YFM shorthand-table cell is not recognised as a fence closer
Bug 30: blockquote-wrapped fence created phantom ranges and swallowed real includes (regression of Bug 29)
Bug 31: empty-resolving include silently drops the next sibling include (different root cause, separate from the fence scanner)
Bug 32: non-ASCII (cyrillic) anchor id makes `extractSection` return the whole file
Bug 33: `cutHeading` ended a `#hash` section at a shallower heading, dropping `notitle` includes


src/commands/build/features/output-md/index.ts
```diff
@@ -226,6 +226,26 @@ export class OutputMd {
...
+                                // dependency concurrently (-j2); without this, the include dump
+                                // could read `meta` before the entry hook populated `vcsPath`,
+                                // producing non-deterministic frontmatter. `vcs.metadata` is
+                                // idempotent (memoized, config gated, deterministic realpath),
+                                // so resolving it here makes the include dump self-sufficient
+                                // and write order irrelevant. Pure includes (not TOC entries)
+                                // are skipped so their output keeps matching the entry path.
+                                if (run.toc.isEntry(graph.path)) {
+                                    const vcsMeta = await run.vcs.metadata(
+                                        graph.path,
```
martyanovandrey, line 242: It duplicates the logic from features/contributors/index.ts and has no tests

martyanovandrey (changes requested):

martyanovandrey (approved):

PR #1966, opened 2026-05-31
feat(themer): add code highlight theme configuration and tests

m-xim:
## Description

Added support for configurable code highlighting themes.

Now documentation can use different highlight.js themes for light and dark UI modes through `.yfm`:

```yaml
codeHighlight:
  light: github
  dark: github-dark
```

### theme.yaml

Existing `theme.yaml` color overrides continue to work with the selected code highlight themes.

For example:

```yaml
light:
  code: '#111111'
  code-background: '#ffeeaa'

dark:
  code: '#eeeeee'
  code-background: '#221144'
```

These values are applied on top of the selected highlight.js theme for `.hljs`, so users can select a full code theme and still override supported Diplodoc code colors.


src/commands/build/features/themer/code-highlight.ts
```diff
@@ -0,0 +1,120 @@
+import type {Run} from '~/commands/build';
+import type {CodeHighlightConfig, ColorVariant, ThemeColors, ThemeConfig} from './types';
+
+import {join} from 'node:path';
+import postcss from 'postcss';
```
martyanovandrey, line 5: Pulling in `postcss` for the current task feels like overkill. I think we can reuse `parseCSS` from `@gravity-ui/uikit-themer` instead — it’s already imported in `utils.ts`.

src/commands/build/types.ts
```diff
@@ -91,6 +91,7 @@ type BaseConfig = {
      * - 'random' (default): uses Math.random() — legacy behavior.
      * - 'deterministic': uses per-file counters with prefix (e.g. 'term-1').
      */
+    codeHighlight?: CodeHighlightConfig | null;
```
martyanovandrey: The JSDoc comment is detached from his idGenerator field

tests/package-lock.json
martyanovandrey, line 1: a lot of diff that doesn't belong to PR

src/commands/build/features/themer/types.ts
```diff
@@ -4,8 +4,14 @@ export type ThemerArgs = {
...
 };
 
+export type CodeHighlightConfig = {
+    light?: string;
+    dark?: string;
+};
+
 export type ThemerConfig = {
     theme?: string | null;
+    codeHighlight?: CodeHighlightConfig | null;
```
martyanovandrey, line 14: CodeHighlightConfig is declared with light?: string and dark?: string, but in code-highlight.ts, the hasCodeHighlight function also accepts null | undefined. However, in ThemerConfig.codeHighlight?: CodeHighlightConfig | null, null is allowed. If a user writes codeHighlight: null in .yfm (intentionally disabling it), but the JSON schema in schemas/yfm-schema.yaml only allows type: object, null will fail validation.

src/commands/build/run.ts
```diff
@@ -86,6 +87,10 @@ export class Run extends BaseRun<BuildConfig> {
         return require(join(__dirname, 'manifest.json')) as Manifest;
     }
 
+    get highlightStylesPath() {
+        return this.realpathSync(HIGHLIGHT_STYLES_ROOT as AbsolutePath);
```
martyanovandrey: The highlightStylesPath getter triggers realpathSync on every access. It is currently used ~once per build, but as a getter in Run, it is potentially called from multiple locations. It is better to calculate it once in the constructor and store it in a field (as is done for originalInput/output).

martyanovandrey (changes requested):

src/commands/build/features/themer/code-highlight.ts
m-xim: I checked `parseCSS` from `@gravity-ui/uikit-themer`.

It doesn't seem to fit this case. It parses Gravity theme CSS into a `GravityTheme` structure, while here we need to process highlight.js CSS and prefix its selectors with the theme scope.

So `parseCSS` doesn't give us the model we need for this transformation.

tests/package-lock.json
```diff
@@ -41,6 +41,7 @@
         "@diplodoc/transform": "^4.72.0",
         "@diplodoc/translation": "^1.7.19",
         "@gravity-ui/uikit-themer": "^1.7.0",
+        "@inquirer/prompts": "^8.3.2",
```
martyanovandrey: What is this/why?

src/commands/build/features/themer/code-highlight.ts
```diff
@@ -0,0 +1,120 @@
...
+        throw new Error(`Unknown highlight.js theme: "${themeName}"`);
+    }
+
+    return themePath;
+}
+
+function scopeCss(css: string, scope: string): string {
+    const root = postcss.parse(css);
+
+    root.walkRules((rule) => {
```
martyanovandrey, line 48: walkRules visits rules inside @keyframes too, so keyframe selectors (from/to/50%) get prefixed and become invalid


@keyframes blink {
    .g-root_theme_light .dc-doc-page .yfm from { opacity: 1; } 
}

Standard hljs themes are color-only, so it doesn't show today, but the feature allows any theme/custom CSS. Maybe smth like:

```
root.walkRules((rule) => {
    const parent = rule.parent;
    if (parent?.type === 'atrule' && /keyframes$/i.test(parent.name)) {
        return;
    }
    if (rule.selectors?.length) {
        rule.selectors = rule.selectors.map((s) => `${scope} ${s}`);
    }
});
```

?

martyanovandrey (approved):

PR #1986, opened 2026-06-08
feat(build): generate llms.txt and llms-full.txt

vgvoleg:
## What

Adds an opt-in `--llms` build feature that emits, per toc, two artifacts
following the [llmstxt.org](https://llmstxt.org) spec:

- **`llms.txt`** — a compact index: `# <toc title>` + per-entry links with
  frontmatter descriptions.
- **`llms-full.txt`** — the whole doc as one self-contained markdown file.

Works for both `md` and `html` output.

## Design

- Runs in `AfterAnyRun` over the **already-filtered** toc, so the artifacts
  match the single-source "version" being built (conditional / hidden pages
  don't leak in).
- Index links point to the real output file: original href for `md`,
  `.html` for `html`.
- `llms-full.txt` is assembled as **fully self-contained markdown** (all
  includes merged), identical for md and html.

To avoid an md/html fork and keep one source of truth, `OutputMd`'s per-page
assembly was extracted into a reusable `MarkdownCollector`
(`output-md/collect.ts`). `OutputMd` delegates to it; `Llms` reuses it with a
`SELF_CONTAINED` preset.

`llms.txt` / `llms-full.txt` are excluded from `build-content-map` content
hashes.

### DIFF
--- a/src/commands/build/features/llms/config.ts
+++ b/src/commands/build/features/llms/config.ts
@@ -1,0 +1,12 @@
+import {option} from '~/core/config';
+
+const llms = option({
+    flags: '--llms',
+    desc:
+        'Generate llms.txt (index) and llms-full.txt (full markdown) per toc for LLM consumption. ' +
+        'Works for both `md` and `html` output.',
+});
+
+export const options = {
+    llms,
+};
--- a/src/commands/build/features/build-content-map/index.ts
+++ b/src/commands/build/features/build-content-map/index.ts
@@ -134,11 +134,20 @@ const ARTIFACT_FILENAMES = new Set([
 // `yfm-redirects-meta-file.json`).
 const META_ARTIFACT_RE = /^yfm-.+-meta.*\.json$/;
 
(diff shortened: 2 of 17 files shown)


src/commands/build/features/llms/config.ts
```diff
@@ -0,0 +1,12 @@
+import {option} from '~/core/config';
+
+const llms = option({
+    flags: '--llms',
+    desc:
+        'Generate llms.txt (index) and llms-full.txt (full markdown) per toc for LLM consumption. ' +
+        'Has effect only for the `md` output format.',
```
martyanovandrey: but it also be generated for the html build?

martyanovandrey (approved):

src/commands/build/features/llms/config.ts
vgvoleg: oh sorry, yes, forgot to update this desc, will fix

vgvoleg: fixed

PR #2003, opened 2026-06-18
feat: add openapi companion DOCSTOOLS-5791

goldserg:
- config section `ai.openapiCompanions` (`true | 'md' | false`, default `'md'`);
- CLI flag `--ai-openapi-companions` (boolean override);
- option `--max-openapi-include-inline-size` (default 100K, max 1M);
- yfm-schema: section `ai` + `content.maxOpenapiIncludeInlineSize` (+ documented
  `content.maxOpenapiIncludeSize`).
- CLI-wrapper openapi writes `*.openapi.json` as is (without stub-logic) and registers the entry;
- build-manifest adds an array of `openapiCompanions` (`{leadingPage, companionPath}`),
  dedup + deterministic sorting;
- HTML-comment with a link to `.openapi.json` for **static** `md2html`-build:
  the viewer adds it when returning only for md2md, but the `md2html` viewer doesn't have it, so the comment
  babbles directly in the `<body>` of a ready-made html-rendering (HTML comments from markdown
 are cut out by the transformation). It is gated by the fact of companion emission (depends on `renderMode`
 and `ai.openapiCompanions`).

### DIFF
--- a/src/commands/build/features/build-manifest/index.ts
+++ b/src/commands/build/features/build-manifest/index.ts
@@ -1,5 +1,5 @@
 import type {Command} from '~/core/config';
-import type {Build, Run} from '~/commands/build';
+import type {Build, OpenapiCompanionEntry, Run} from '~/commands/build';
 import type {Redirects} from '../../services/redirects';
 
 import {join, parse} from 'node:path';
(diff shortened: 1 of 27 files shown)


src/commands/build/features/build-manifest/index.ts
```diff
@@ -29,10 +29,19 @@ type FileTrieEntryPoint = {
     tocMapping: Record<string, string>;
 };
 
+/** Maps a generated OpenAPI leading page to its standalone spec companion file. */
+type OpenapiCompanionEntry = {
```
reazy015: It looks like this type is duplicated — the same one we have in the OpenAPI extension and the output-html feature. Is it necessary to have three separate types?

reazy015 (approved):

PR #2004, opened 2026-06-19
feat: removed schemas and replaced it with @diplodoc/ajv

separatrixxx:
The schemes from the package have been moved to https://github.com/diplodoc-platform/ajv


tests/mocks/redirects-validation/unparseable/input/redirects.yaml
```diff
@@ -1,11 +1,4 @@
...
     to: /bar.md
-
-<<<<<<< HEAD
-
-  - from: /barbar.md
-    to: /baz.md
-
-=======
->>>>>>> merge-branch
+
```
goldserg: why is that?
It's a special test

goldserg (approved):

PR #2013, opened 2026-06-23
fix(build-content-map): stabilize content hash for inlined SVG ids

martyanovandrey:
## Problem

The documentation changes digest reports pages as “changed” even though their Markdown source did not change between revisions. In a real run, 5 out of 6 pages were false positives.

The diff is built from the `yfm-build-content.json` artifact via the `build-content-map` feature: a page is added to `changed` if its content hash differs between revisions.

## Cause

In md2md, inline SVG is processed through SVGO `prefixIds`, which rewrites all IDs inside the SVG to `svg-<random>__<id>`. With the default `--id-generator random` strategy, the prefix is generated using `Math.random()` and changes on every build, even when the source is unchanged. As a result, the page content hash differs on every rebuild → false positives in the diff.

This is the only volatile pattern in md2md output: terms, tabs, and inline code remain in Markdown syntax and do not receive IDs at this stage.

## Solution

I’m extending the existing `normalizeForHash` logic — the same place where `updatedAt`, `contributors`, and `metadata.generator` are stripped. Before hashing, the volatile prefix `svg-<token>__` is canonicalized to `svg__`. This covers both `id="…"` and `url(#…)`/`href` references: everything is rewritten consistently, so links remain valid. The actual SVG content — paths, dimensions, and the original IDs after `__` — is preserved, so real changes still affect the hash.

This is a surgical fix: it stabilizes only the hash in `yfm-build-content.json` and **does not change the build output or defaults**.

## Verification

- Building the same revision twice with the default `random` strategy: content hashes for all unchanged pages are **identical**, while the raw output bytes still differ because the IDs remain random. Pages with inline SVG included via `include` are stable as well.
- Sensitivity to real SVG edits is preserved.
- Unit tests, including 5 new canonicalization tests, and the `build-content-map` e2e test are green.
- Build time impact: `.replace()` is about 4× cheaper than SHA-256 on the same content (~10 GB/s); on documentation with 1000 pages, it adds ~0.8 ms.

### DIFF
--- a/src/commands/build/features/build-content-map/index.ts
+++ b/src/commands/build/features/build-content-map/index.ts
@@ -234,8 +234,36 @@ function hasVolatile(obj: Record<string, unknown>): boolean {
     return false;
 }
 
+// The inline-SVG transform (md2md, see output-md/plugins/merge-svg.ts) inlines
+// SVG assets and runs them through SVGO `prefixIds`, which rewrites every id
+// inside the SVG to `<prefix>__<id>`. The prefix comes from the build's
+// id-generator; with the default `random` strategy it is `svg-<8 base36 chars>`
+// derived from `Math.random()`, so it differs on every build of an unchanged
+// source. That volatile prefix would flip the page's content-hash on each
+// rebuild, producing false-positive diffs for consumers (change
+// notifications, search reindex). Canonicalize it to a fixed marker so the
+// hash reflects the SVG's content, not its per-build ids. The same prefix
+// appears in both `id="…"` definitions and their `url(#…)` / href references,
+// and the global replace rewrites all of them identically, so they stay
+// internally consistent. Real SVG content (paths, dimensions, the original
+// id stems after `__`) is preserved, so genuine changes still move the hash.
+//
+// Coupled to transform's prefix shape (`@diplodoc/transform`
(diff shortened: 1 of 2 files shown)


src/commands/build/features/build-content-map/index.ts
```diff
@@ -246,25 +274,25 @@ function normalizeMarkdown(content: Buffer): Buffer {
         body = stripped;
         rawFrontmatter = raw;
     } catch {
-        // Malformed YAML inside the fences — hash the raw bytes.
-        return content;
+        // Malformed YAML inside the fences — hash the SVG-canonicalized bytes.
+        return Buffer.from(text, 'utf8');
```
goldserg, line 278: And why is that?

goldserg (approved):

src/commands/build/features/build-content-map/index.ts
martyanovandrey: text is already svg canonicalized (this pass goes before parsing frontmatter). Broken YAML prevents you from cutting updatedAt/generator, but you still need to normalize the body with embedded SVG, otherwise the hash of such a page will remain unstable, i.e. the very false-positive that we fix. Therefore, I return the canonicalized text instead of the raw content.

PR #2017, opened 2026-06-25
fix: exclude build dir from repeated input

reazy015:


src/core/run/index.ts
```diff
@@ -177,6 +177,13 @@ export class Run<TConfig = BaseConfig> {
             return [];
         }
 
+        if (!isFile) {
+            const rel = relative(from, to);
+            if (rel && !rel.startsWith('..') && !isAbsolute(rel)) {
+                ignore = [...ignore, `${rel}/**`];
+            }
```
goldserg: duplicate, may be use core/utils for this code?

goldserg (approved):

PR #2022, opened 2026-06-26
fix: yaml compose format fix

reazy015:
#### Use yaml stringify instead of js-yaml dump

The main reason is that js-yaml dump strips some unicode symbols, trailing space in strict mode and thus we have differently formated text on compose output procedure.

Yaml stringify method acts differently and may save initial formatting, except one edge case related to trailing space in multiline string

### DIFF
--- a/src/commands/translate/utils/fs.ts
+++ b/src/commands/translate/utils/fs.ts
@@ -2,7 +2,8 @@ import type {AjvOptions, JSONObject, LinkedJSONObject} from '@diplodoc/translati
 
 import {dirname, resolve} from 'node:path';
 import {mkdir, readFile, writeFile} from 'node:fs/promises';
-import {dump, load} from 'js-yaml';
+import {load} from 'js-yaml';
+import {stringify} from 'yaml';
 import {linkRefs, unlinkRefs} from '@diplodoc/translation';
 import {
     leadingSchemaJson,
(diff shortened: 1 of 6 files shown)


src/commands/translate/utils/fs.ts
```diff
@@ -43,7 +44,20 @@ function stringifyFile(content: JSONObject | string, path: string): string {
 
     switch (ext(path)) {
         case 'yaml':
-            return dump(content, {noRefs: true, lineWidth: -1});
+            return stringify(
+                content as Record<string, unknown>,
+                (_key, value) => {
+                    if (typeof value === 'string') {
+                        return value.replace(/[ \t]+$/, '');
```
stenin-nikita, line 51: Can you tell me what this logic is for here? Also, it seems better to use `value.trimEnd()` in that case

src/commands/translate/utils/fs.ts
```diff
@@ -2,7 +2,8 @@ import type {AjvOptions, JSONObject, LinkedJSONObject} from '@diplodoc/translati
 
 import {dirname, resolve} from 'node:path';
 import {mkdir, readFile, writeFile} from 'node:fs/promises';
-import {dump, load} from 'js-yaml';
+import {load} from 'js-yaml';
+import {stringify} from 'yaml';
```
stenin-nikita, line 6: Maybe it's worth planning to migrate to one solution instead of two. Should we create a ticket for this?

stenin-nikita (approved):

PR #2030, opened 2026-06-29
feat: added support for documentation with ya.make

separatrixxx:


src/commands/build/features/ya-make/config.ts
```diff
@@ -0,0 +1,17 @@
+import {resolve} from 'node:path';
+
+import {option} from '~/core/config';
+
+const yaMake = option({
+    flags: '--ya-make <path>',
+    desc: `
+        Path to Arcadia root. Enables ya.make support.
```
goldserg, line 8: Is this the path to the root of Arcadia, or to the ya.make file?

separatrixxx: to the root of arcadia

goldserg (approved):

PR #2044, opened 2026-07-08
fix: add restricted-access to build-manifest DOCSTOOLS-6180

goldserg:


src/commands/build/features/build-manifest/index.ts
```diff
@@ -29,11 +29,14 @@ type FileTrieEntryPoint = {
     tocMapping: Record<string, string>;
 };
 
+type RestrictedAccessMap = Record<string, string[][]>;
```
martyanovandrey, line 32: correct the scheme

martyanovandrey (approved):

PR #2077, opened 2026-07-17
fix: add resolve llms enable

HelenVirtanen:
## Description

Adds configurable enable/disable logic for llms.txt / llms-full.txt generation, resolving how the --llms CLI flag interacts with the llms.enabled config option and the build's output format.

Previously, enabling generation was a simple boolean OR between the CLI flag and the config value, with no way to:
explicitly disable the feature via --no-llms when it's enabled in the config,
distinguish "no config for llms at all" from "explicitly disabled in config",
scope config-driven enablement to md2md builds only, while still allowing HTML output when the user opts in explicitly (via flag or config).
This PR introduces a clear, tri-state resolution model.

Behavior

--llms flag: not passed | config.llms.enabled: no llms section (undefined) | Output format: md | Result: generated
--llms flag: not passed | config.llms.enabled: no llms section (undefined) | Output format: html | Result: skipped
--llms flag: not passed | config.llms.enabled: true | Output format: md or html | Result: generated
--llms flag: not passed | config.llms.enabled: false | Output format: md or html | Result: skipped
--llms flag: --llms | config.llms.enabled: any | Output format: md or html | Result: generated
--llms flag: --no-llms | config.llms.enabled: any | Output format: md or html | Result: skipped

Rules, in order of precedence:

1) Explicit --llms always wins - generation runs for any output format, regardless of config.
2) Explicit --no-llms always wins - generation never runs, regardless of config.
3) If the flag isn't passed at all:
llms.enabled: false in config -> disabled.
llms.enabled: true in config -> enabled for any format (treated with the same weight as an explicit --llms).
4) No llms section in config at all -> enabled only for md2md builds (the sensible default without any opt-in).


src/commands/build/features/llms/index.ts
```diff
@@ -60,13 +60,14 @@ export class Llms {
         });
 
         getBaseHooks(program).Config.tap('Llms', (config, args) => {
-            const llmsArg = args.llms || false;
-            const llmsEnabled = config?.llms?.enabled || false;
+            const configLlmsEnabled: boolean | undefined = config?.llms?.enabled;
             const llmsDescription = config?.llms?.description || '';
+            const onlyMd = config.outputFormat === OutputFormat.md;
```
goldserg, line 65: Try the approach here
https://github.com/diplodoc-platform/cli/blob/master/src/commands/build/features/output-md/index.ts#L62-L105

goldserg (approved):

PR #2103, opened 2026-07-22
fix: Clear llms-full from style, script DOCSTOOLS-6205

goldserg:


tests/e2e/__snapshots__/merge-includes.spec.ts.snap
```diff
@@ -1709,12 +1703,7 @@ Some description of option A.
 
 [Subscribe](https://example.com/subscribe){.button}
 
-<style>
-a.button {
-  color: black;
-  background: yellow;
-}
-</style>
+
```
martyanovandrey, line 1706: Why are these snapshots changing?

src/commands/build/features/llms/utils.ts
```diff
@@ -0,0 +1,73 @@
+/**
```
martyanovandrey, line 1: if there is a space before ```html, will it break?

src/commands/build/features/llms/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+ * Removes HTML tags (with content) from markdown text.
```
martyanovandrey, line 2: why regexps and not tokens? what's the difference in performance?

tests/e2e/__snapshots__/merge-includes.spec.ts.snap
goldserg: We remove these tags from llms-full, and there are no changes for md

martyanovandrey (approved):

PR #2135, opened 2026-07-30
feat(translate): persistent translation cache for AI providers

martyanovandrey:
## Summary

Follow-up to #1906. Repeated `translate` runs used to retranslate the whole doc set: tokens were spent again and already reviewed translations drifted between runs because LLM output is not deterministic. This is the standard translation-memory problem; TMS products and doc-translation pipelines solve it with a per-segment cache.

### What's added

- `--cache-dir <path>` (or `cacheDir` in config) enables a file-backed translation memory: unit hash -> translation, one JSON file per provider and language pair (`<provider>.<source>-<target>.json`).
- Units found in the store never reach the LLM; the run stat line reports `cached-units: N`.
- The cache fingerprint covers provider, model, prompts, prompt mode and glossary - changing any of them resets stored translations instead of serving stale ones.
- `--no-cache` disables the store for a single run.
- The store is flushed after every processed file, so an interrupted run keeps its progress.
- `--dry-run` reads the cache (estimates reflect the real remaining spend) but never writes it.

The in-memory in-flight dedupe from #1906 stays as is; the store sits in front of it.

DOCSTOOLS-6386

### DIFF
--- a/src/commands/translate/providers/ai/provider.ts
+++ b/src/commands/translate/providers/ai/provider.ts
@@ -340,7 +340,7 @@ type TranslatorParams = {
  * The fingerprint covers everything that affects the output, so changing
  * the model, prompts or glossary safely resets the cache.
  */
-function makeStore(
+export function makeStore(
     client: LLMClient,
     config: AITranslationConfig,
     sourceLanguage: string,
--- a/src/commands/translate/providers/ai/provider.ts
+++ b/src/commands/translate/providers/ai/provider.ts
@@ -23,12 +23,7 @@ import {
     cacheFingerprint,
     estimateTokens,
 } from './utils';
-import {
-    DEFAULT_SYSTEM_PROMPT,
-    DEFAULT_USER_PROMPT,
-    buildMessages,
-    splitFragments,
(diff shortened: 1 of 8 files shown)


src/commands/translate/providers/ai/provider.ts
```diff
@@ -184,12 +201,49 @@ type TranslatorParams = {
...
+    config: AITranslationConfig,
+    sourceLanguage: string,
+    targetLanguage: string,
+): TranslationStore | undefined {
+    if (!config.cacheDir) {
+        return undefined;
+    }
+
+    const file = join(config.cacheDir, `${client.name}.${sourceLanguage}-${targetLanguage}.json`);
+    const fingerprint = cacheFingerprint({
```
Yurok868, line 356: We should include DEFAULT_SYSTEM_PROMPT and DEFAULT_USER_PROMPT in the cache fingerprint.

https://github.com/diplodoc-platform/cli/blob/master/src/commands/translate/providers/ai/prompts.ts#L21

Yurok868 (approved):

PR #2136, opened 2026-07-30
feat(translate): optional LLM-as-judge quality report for AI translations

martyanovandrey:
## Summary

Follow-up to #1906. Adds an opt-in quality evaluation pass over AI translations (the LLM-as-judge pattern used by GenAIScript and Crowdin AI Pipelines): it tells the team what to send to human review instead of proofreading everything.

### How it works

- `--judge` enables the pass; after all files of a target language are translated, source/translation pairs are scored 0-100 by a judge model.
- `--judge-model <name>` picks a different (e.g. stronger) model for scoring; defaults to the translation model.
- Segments scoring below `--judge-threshold` (default 70) are reported as per-file warnings and collected into `translate-quality.<target>.json` in the output directory (score, short issue note, source and translation).

### Guardrails

- Strictly opt-in: scoring roughly doubles token usage.
- Best-effort by design: identical pairs are deduped before scoring, unparsable or failed judge batches are logged and skipped - quality evaluation never fails the translation run.
- Untranslated units (output identical to source) are not scored.
- `--dry-run` skips the judge pass entirely.

DOCSTOOLS-6387


src/commands/translate/providers/ai/provider.ts
```diff
@@ -106,6 +126,69 @@ export class Provider {
...
+                verdict.source.length > 80 ? verdict.source.slice(0, 80) + '...' : verdict.source;
+            this.logger.warn(
+                verdict.path,
+                `Translation quality ${verdict.score}/100: ${preview}` +
+                    (verdict.issue ? ` (${verdict.issue})` : ''),
+            );
+        }
+
+        const report = join(resolve(config.output), `translate-quality.${targetLanguage}.json`);
+        writeFileSync(
```
stenin-nikita: Is writeFileSync intentional here, or could we switch to fs.promises.writeFile to avoid blocking the event loop?

src/commands/translate/providers/ai/provider.ts
```diff
@@ -106,6 +126,69 @@ export class Provider {
...
+        });
+
+        const threshold = config.judgeThreshold;
+        const low = verdicts
+            .filter((verdict) => verdict.score < threshold)
+            .sort((a, b) => a.score - b.score);
+
+        for (const verdict of low) {
+            const preview =
+                verdict.source.length > 80 ? verdict.source.slice(0, 80) + '...' : verdict.source;
```
stenin-nikita: What does 80 represent here? It looks like a magic number — might be worth extracting it into a named constant for readability and easier maintenance.

stenin-nikita (approved):

PR #2158, opened 2026-08-05
feat: added page tags

separatrixxx:


src/extensions/local-search/provider.ts
```diff
@@ -86,17 +86,19 @@ export class LocalSearchProvider implements SearchProvider {
...
         const link = this.pageLink(lang);
         const prettyLink = skipHtmlExtension ? shortLink(link) : link;
+        const tags = includeTags ? this.indexer.getTags(lang) : [];
 
         return {
             ...this._config,
             api: this.apiLink,
             provider: 'local',
             link: prettyLink,
+            ...(tags.length && {tags}),
```
stenin-nikita: Why is `{...(tags.length && {tags})}` used here instead of simply `{tags}`? Is it important to omit the tags field when the array is empty?

src/core/template/template.ts
```diff
@@ -388,6 +416,14 @@ function meta(record: Hash<string>) {
     return `<meta ${attributes(record)}>`;
 }
 
+function escapeAttributeValue(value: string) {
+    return value
+        .replace(/&/g, '&amp;')
+        .replace(/"/g, '&quot;')
+        .replace(/</g, '&lt;')
+        .replace(/>/g, '&gt;');
+}
```
stenin-nikita: I think the project should already have a utility for escaping HTML attribute values. Could we reuse it instead of adding a new function?

stenin-nikita (approved):

PR #2168, opened 2026-08-07
fix(translate): restore missing <source> wrapper in LLM responses

martyanovandrey:
## Problem

The LLM occasionally returns a translated unit without the `<source xml:space="preserve">...</source>` wrapper. The raw response is resolved to concurrent files and persisted into the translation cache as is. Compose is strict about the wrapper (`Did not find any translations` assert in @diplodoc/translation), so a single such unit fails the whole file with COMPOSE_ERROR, and the poisoned cache entry keeps failing every subsequent run that hits it.

Observed on a real run against deepseek-v4-flash: 1583 of 5996 cached units were stored without the wrapper, which failed compose for 92 of 108 files.

## Fix

Normalize the response before it reaches the in-run cache and the persistent store: if the source unit is wrapped and the response is not, restore the wrapper.

### DIFF
--- a/src/commands/translate/providers/ai/provider.spec.ts
+++ b/src/commands/translate/providers/ai/provider.spec.ts
@@ -224,6 +224,22 @@ describe('translate ai provider', () => {
             expect(client.complete).toHaveBeenCalledTimes(1);
         });
 
+        it('should restore the <source> wrapper when the model strips it', async () => {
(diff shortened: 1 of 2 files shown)


martyanovandrey:
Superseded by #2159: the provider now strips the <source> wrapper before prompting and unconditionally restores it before the cache/store write, so this conditional re-wrap can never fire.

PR #2190, opened 2026-08-13
fix: improved tag normalization

separatrixxx:


src/core/meta/MetaService.ts
```diff
@@ -107,7 +112,27 @@ export class MetaService {
...
+        }
+
+        if (meta.tags) {
+            meta.tags = [
+                ...new Set(
+                    meta.tags.map((tag) => {
+                        const trimmedTag = tag.trim();
+                        const normalizedTag = Array.from(trimmedTag.toLowerCase());
+
+                        if (Array.from(trimmedTag).length > MAX_TAG_LENGTH) {
```
martyanovandrey: The warn condition and the slice operate on different strings: the length check uses the pre-lowercase string, but `slice(0, MAX_TAG_LENGTH)` is applied to the lowercased array. Some characters expand on lowercasing - e.g. Turkish `İ` (U+0130) becomes two code points (`i` + combining dot). So for `'İ'.repeat(32)` the check sees 32 code points and stays silent, while the slice cuts 64 -> 32, silently dropping half of the tag.

Warn should fire iff truncation actually happens - measure the same array you slice:

```suggestion
                        if (normalizedTag.length > MAX_TAG_LENGTH) {
```

src/core/meta/MetaService.spec.ts
```diff
@@ -52,15 +53,61 @@ describe('MetaService', () => {
...
+            const file = 'test/file.md' as NormalizedPath;
+
+            metaService.add(file, {
+                tags: [
+                    ' info ',
+                    '_Internal',
+                    '',
+                    'info',
+                    'Tag',
+                    'tag',
```
martyanovandrey, line 74: This test enshrines the silent-truncation case: it asserts no warning for `'İ'.repeat(32)`, but the resulting tag IS truncated (lowercase expands it to 64 code points, slice cuts it back to 32). With the condition fixed to measure `normalizedTag`, this expectation should flip to `toHaveBeenCalled()` - and it's worth also asserting the dumped value so the truncation itself is visible in the test.

martyanovandrey (approved):

PR #2205, opened 2026-08-18
feat(translate): support --context-file for AI providers DOCSTOOLS-6493

martyanovandrey:
## Overview

`yfm translate` now accepts arbitrary reference materials for the translation prompt: project descriptions, style guides, UI text lists, terminology tables and other context of arbitrary structure. Real translation pipelines carry glossaries as markdown tables with definitions and free-form style guides - none of which fit the strict `--glossary` YAML pairs.

## Details

- New repeatable `--context-file <value>` option (config alternative: `contextFiles` list in the yfm config). Accepts a path to a text file (md, json, txt - the content is passed to the model as-is) or a literal multi-line text block.
- By default sections are appended to the system prompt: they are identical for every batch, and a static system prompt plays well with provider-side prompt caching. A `{{contextFiles}}` placeholder in `--system-prompt` or `--user-prompt` overrides the placement.
- Unlike prompts, a single-line value that does not resolve to an existing file is a configuration error, not a literal: a typo in a path must not silently become prompt text.
- Context content participates in the translation cache fingerprint, so editing a context file safely resets the cache. The fingerprint stays unchanged when no context files are configured, existing caches survive the update.

## Config usage

The main consumer is CI, where translation parameters live in the docs repo rather than in the pipeline: the `translate` section of the `.yfm` at the input root is picked up automatically, and relative paths are resolved from the `.yfm` location.

```yaml
translate:
  contextFiles:
    - ./neurotranslate/project_info.md
    - ./neurotranslate/glossary.md
    - ./neurotranslate/ui_elements_list.md
```

Credentials stay out of the config by design: `--auth` is accepted only as a flag or environment variable.

### DIFF
--- a/src/commands/translate/providers/ai/prompts.ts
+++ b/src/commands/translate/providers/ai/prompts.ts
@@ -1,5 +1,6 @@
 import type {ChatMessage} from './clients/types';
 
+import {ok} from 'node:assert';
 import {existsSync, readFileSync} from 'node:fs';
 import {dedent} from 'ts-dedent';
 
@@ -16,6 +17,8 @@ export type PromptConfig = {
     glossaryPairs: GlossaryPair[];
     /** Human-readable document context, e.g. `document "Quickstart" (file docs/ru/index.md)`. */
     context?: string;
+    /** Resolved contents of --context-file values, injected as reference material. */
+    contextFiles?: string[];
 };
 
 const FRAGMENT_SEPARATOR = '<<<§§§>>>';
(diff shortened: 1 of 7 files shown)


src/commands/translate/providers/ai/prompts.ts
```diff
@@ -77,6 +80,43 @@ export function resolvePromptValue(
...
+ * Resolves a --context-file value: a path to a text file or a literal
+ * multi-line text block. Unlike prompts, a single-line value is always
+ * a path, so a missing file is a configuration error, not a literal.
+ */
+export function resolveContextValue(value: string, resolve?: (path: string) => string): string {
+    const trimmed = value.trim();
+
+    // A real file path never contains a line break.
+    if (!trimmed.includes('\n')) {
+        if (existsSync(trimmed)) {
```
Yurok868: if resolve is provided, should we use it first?

```
const path = resolve ? resolve(trimmed) : trimmed;

if (existsSync(path)) {
    return readFileSync(path, 'utf8');
}

ok(false, `Context file not found: ${value}`);
```

Yurok868 (approved):

src/commands/translate/providers/ai/prompts.ts
martyanovandrey: Good point - made it exclusive: config values now resolve from the config dir only

PR #2216, opened 2026-08-21
fix: sync meta from worker thread to main thread for post-processing hooks

goldserg:
When --jobs is enabled, process() runs in a worker thread with its own MetaService instance. The worker loads frontmatter (including noIndex and restricted-access) into its local MetaService, but this data never reaches the main thread's MetaService.

AfterRun hooks (Llms excludeNoIndex, BuildManifest collectRestrictedAccess) call run.meta.dump(path) in the main thread and get empty meta, causing noIndex pages to leak into llms.txt and restricted-access rules to vanish from the build manifest.

Fix: process() now returns result.data.meta alongside result.info, and processEntry() writes it back into the main thread's MetaService via run.meta.set(entry, info.meta).

### DIFF
--- a/src/commands/build/features/llms/index.ts
+++ b/src/commands/build/features/llms/index.ts
@@ -3,6 +3,7 @@ import type {Command} from '~/core/config';
 import type {Toc} from '~/core/toc';
 
 import {dirname, join, relative} from 'node:path';
+import {extractFrontMatter} from '@diplodoc/liquid';
 
 import {defined} from '~/core/config';
 import {getHooks as getBaseHooks} from '~/core/program';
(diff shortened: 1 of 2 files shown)


src/commands/build/features/llms/index.ts
martyanovandrey, line 1: The llms fix itself looks good, but the PR description says the bug also breaks restricted-access in the build manifest - and that part isn't fixed: collectRestrictedAccess still calls run.meta.dump() from the main thread, so with --jobs it will keep getting empty meta.

The description also mentions a different fix (returning meta from process() and syncing it back via run.meta.set()), which would cover both places at once. Was there a reason to drop that approach? If yes, let's file a follow-up for build-manifest and update the description to match the actual code.

martyanovandrey (approved):

PR #2220, opened 2026-08-25
fix(translate/seed): tolerate per-file load and extraction errors

martyanovandrey:
A single file whose source or target fails to load or extract (broken page-constructor YAML, frontmatter quirks) crashed the whole `translate seed` run: no stats line, no seed file written, and exit code 0, so callers could not even detect the failure.

Reproduced on a real 713-file project: one unparseable release-notes target left the project completely unseeded.

Fix: per-file try/catch - a broken file is reported with a WARN and counted in a new `failed` stats bucket (falling back to a full retranslation, exactly like a unit-count mismatch), while every healthy file is still seeded and the store is flushed. With the fix the same project seeds 388 files / 16527 units, `failed: 10`.

Found while replacing the translation engine of the internal Arcadia tasklet with `yfm translate` (DOCSTOOLS-6368).

### DIFF
--- a/src/commands/translate/commands/seed-failure.spec.ts
+++ b/src/commands/translate/commands/seed-failure.spec.ts
@@ -14,7 +14,7 @@ vi.mock('../utils', async (importOriginal) => {
     return {
         ...original,
         loadTranslationUnits: vi.fn(async (params: {inputPath: string}) => {
-            if (String(params.inputPath).includes('broken')) {
+            if (params.inputPath.includes('broken')) {
                 throw new Error('Unable to extract valid tokens for text segment.');
             }
             return original.loadTranslationUnits(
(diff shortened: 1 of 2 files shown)


src/commands/translate/commands/seed-failure.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+import {SeedStore, seedFilePath} from '../providers/ai/utils';
+
+import {seedTranslations} from './seed';
+
+vi.mock('../utils', async (importOriginal) => {
+    const original = (await importOriginal()) as typeof translateUtils;
+    return {
+        ...original,
+        loadTranslationUnits: vi.fn(async (params: {inputPath: string}) => {
+            if (String(params.inputPath).includes('broken')) {
```
kadymov: Perhaps this is an unnecessary wrapping in String() `params: {inputPath: string}`

martyanovandrey: Right, the parameter is already typed as string - dropped the wrapper.

kadymov (approved):

PR #2225, opened 2026-08-25
feat(translate): add translation quality eval harness

martyanovandrey:
## What

An eval harness for the AI translation pipeline: a fixed corpus of reference pages is run through `yfm translate`, quality is scored automatically, and the run produces a scorecard (human-readable + JSON) with a pass/fail verdict. Intended to be run before changing translation prompts or models and before rolling the tooling out.

One command:

```bash
npm run translate:eval
```

The default mode needs no network and no API keys: the corpus reference translations are served by a local OpenAI-compatible mock endpoint through the real pipeline (extract, batching, prompts, compose, judge). With credentials, `--real --provider ... --model ...` evaluates an actual provider on the same corpus. Docs: `docs/translate-eval.md`.

## What is checked

- Deterministic markup preservation (source vs translation): code fences byte-identical, liquid/YFM directives (notes, cuts, tabs, includes, conditions) keep order and parameters, link and image targets, heading structure and anchors, `{{variables}}`, table layout.
- Glossary compliance: every glossary term used on a source page must be rendered with its required translation (`tests/eval/corpus/glossary.yaml`, also passed to `--glossary`).
- Untranslated or lost segments: no source-script text outside code fences, except lines the reference keeps in the source language on purpose; missing output files are reported.
- Reference similarity: token-level F1 against the reference translation (trend metric, gated only with `--min-similarity`).
- LLM judge: reuses the existing `--judge` scoring of the translate command; the judge report is folded into the scorecard. In mock mode judge requests are answered deterministically.

Thresholds are configurable (`--max-*`, `--min-judge-score`, `--min-similarity`); the process exits non-zero on FAIL.

## Corpus

`tests/eval/corpus`: 12 pages ru -> en plus 4 shared includes and `toc.yaml`, taken from the real Diplodoc documentation (diplodoc-platform/docs, MIT), which maintains line-aligned ru/en page pairs. Coverage: plain prose, tabs, notes, cuts, code fences, inline code, includes (with anchors), term definitions, liquid variables and `not_var` escaping, links and anchors, lists, markdown pipe tables, YFM grid tables, images. Reference translations live under `en/`, the glossary under `glossary.yaml`. Page list and editing rules: `tests/eval/README.md`.

## Implementation notes

- The harness is built strictly on top of the public CLI surface: it spawns `yfm translate` and talks to it through the documented `--api-base` endpoint. Translate internals are not modified. The only imports from the translate module are the prompt/judge helpers needed to stay in sync with the request format, each covered by a drift-guard test.
- The mock translation memory is built by two capture runs through the real pipeline (`--user-prompt '{{context}}\n{{fragments}}'` against a local echo endpoint), paired positionally per file. `yfm translate seed` could not be used for this: inline element ids in extracted units come from a process-global counter, so hash-keyed seed entries miss across processes. Details in `src/commands/translate/eval/MODULE.md`.
- Harness logic lives in `src/commands/translate/eval/` (typechecked, linted, unit-tested with the regular suite; 59 new unit tests). `scripts/translate-eval.mjs` bundles it with esbuild and runs it. An e2e spec (`tests/e2e/translate-eval.spec.ts`) runs the harness end-to-end on the mock.

## Found along the way

The eval reproduces an upstream compose bug: a live `{% file %}` directive is mangled on the translate round-trip (`name="{% src=... %}"` instead of `name="..."`). The same artifact is visible in the published en docs (`en/syntax/links.md` in diplodoc-platform/docs). The live directive is excluded from the corpus for now; a fenced example remains.


src/commands/translate/eval/markup.ts
```diff
@@ -0,0 +1,230 @@
...
+        gridMarkers += countInline(trimmed, '||');
+    }
+
+    return {
+        fences: page.fences,
+        liquid,
+        links,
+        headings,
+        variables,
+        tables: {pipeRows, gridMarkers},
```
kadymov: The function does not return the number of columns. Or does this not require verification?

src/commands/translate/eval/report.ts
```diff
@@ -0,0 +1,169 @@
...
+            `glossary violations: ${totals.glossary} (allowed: ${thresholds.maxGlossaryViolations})`,
+        );
+    }
+    if (totals.untranslated > thresholds.maxUntranslated) {
+        failures.push(
+            `untranslated lines: ${totals.untranslated} (allowed: ${thresholds.maxUntranslated})`,
+        );
+    }
+
+    if (judge && thresholds.minJudgeScore > 0 && judge.scored > 0) {
```
kadymov: If `skippedPairs > 0`, do we need to add anything to `failures`?

src/commands/translate/eval/glossary.ts
```diff
@@ -0,0 +1,112 @@
...
+        .join('\n');
+    const translatedText = translated.toLowerCase();
+
+    for (const pair of pairs) {
+        const occurrences = countStemOccurrences(sourceText, stemTerm(pair.sourceText));
+        if (!occurrences) {
+            continue;
+        }
+
+        if (!translatedText.includes(pair.translatedText.toLowerCase())) {
```
kadymov, line 57: It seems there might be a case where the pair is found not in the text, but, for example, in a link.
The pair is `link`, and the text in translatedText is `... [documentation](./links.md) ...`, and the condition won’t work.

src/commands/translate/eval/markup.ts
martyanovandrey: Good catch, fixed in e7324c5b: the check now records the pipe count of every table row and compares them as a sequence, so a merged or split column is caught even when the row count stays the same.

src/commands/translate/eval/report.ts
martyanovandrey: Agreed, fixed in e7324c5b: when the judge gate is active, any unscored pairs now fail the run - they were silently weakening the average score.

src/commands/translate/eval/glossary.ts
martyanovandrey: Valid, fixed in e7324c5b: both sides are now reduced to reader-visible text (no fences, inline code, liquid directives or link destinations, link text is kept) before matching, so `link` inside `./links.md` neither counts as usage nor satisfies the required translation. Covered by new specs.

kadymov (approved):

PR #2245, opened 2026-09-08
feat(toc): support inherited noIndex

kadymov:
## Problem

`noIndex` can be set in page frontmatter and global configuration, but `toc.yaml` cannot consistently prohibit indexing for a page or a whole subtree. This also leaves TOC-only restrictions out of LLMS artifacts.

## Changes

- expose boolean `noIndex` on TOC roots and items;
- propagate the effective value monotonically through sections and included TOCs, including a section's own `href`;
- preserve `true` when TOC metadata and frontmatter are merged in either order and loading mode;
- exclude affected pages from `llms.txt` and `llms-full.txt` using the resolved TOC, including worker builds;
- keep `hidden` independent: hidden pages remain absent from LLMS output but do not receive `meta.noIndex`;
- add unit and E2E coverage for roots, pages, sections, siblings, nested `false`, named and expanded includes, metadata conflicts, and `--jobs`.

## Verification

- `npm test` — 1919 tests passed, 1 skipped;
- `npm run typecheck`;
- `npm run lint` — no errors;
- `npm run build`;
- `npm run e2e` — 214 tests passed, 1 skipped, 1 todo;
- focused LLMS E2E — 5 tests passed;
- `git diff --check`.

Schema: [diplodoc-platform/ajv#64](https://github.com/diplodoc-platform/ajv/pull/64)

Task: DOCSTOOLS-4902


src/commands/build/features/llms/index.spec.ts
martyanovandrey, line 1: Looks right for the single-branch case, but the toc flag and the frontmatter flag land in two different places. I tried a toc where shared.md is linked both from a public section and from a section with noIndex: true: the build stamps noIndex: true into shared.md's frontmatter and drops it from the search index, but llms.txt still lists it. collectEntries prunes per branch, applyNoIndex unions per page, so they disagree. run.meta on the main thread already has the toc flag in both --jobs and plain mode. Should excludeNoIndex also OR in run.meta.get(entry.path)?.noIndex, or is the double-linked page considered out of scope here?

src/commands/build/features/llms/index.ts
```diff
@@ -212,8 +213,10 @@ export class Llms {
     private collectEntries(toc: LlmsTocItem, tocDir: string): LlmsEntry[] {
         const entries: LlmsEntry[] = [];
 
-        const visit = (item: LlmsTocItem, parentName = '') => {
-            if (item.hidden) {
+        const visit = (item: LlmsTocItem, parentName = '', inheritedNoIndex = false) => {
```
martyanovandrey: Am I reading this right that inheritedNoIndex can never be anything but false? The guard on 219 returns as soon as noIndex is true, so the recursive call on 233 is only reached when it's false. If so, excludes noIndex pages and descendant pages passes purely on subtree pruning and would still pass with the parameter deleted. Can we drop the third arg and keep just the early return?

src/core/meta/MetaService.ts
```diff
@@ -183,13 +184,21 @@ export class MetaService {
         const file = normalizePath(path);
 
         if (this.config.rawAddMeta) {
+            const current = this.meta.get(file);
+            const noIndex = current?.noIndex === true || record.noIndex === true;
+
             if (isRaw) {
-                this.meta.set(file, record);
+                this.meta.set(file, noIndex ? {...record, noIndex: true} : record);
+            } else if (noIndex) {
```
martyanovandrey, line 192: In rawAddMeta mode current is the same object the caller passed in - MarkdownService keeps it in its load cache as result.meta. So mutating it here writes a noIndex into someone else's frontmatter. Watch mode makes it reachable: markdown is cached, then a toc edit re-runs toc.init (watch/index.ts:327), applyNoIndex fires, and run.markdown.meta(path) starts returning {noIndex: true} for a file with no --- block, which flips isMeaningfulFrontmatter and kills frontmatter propagation. Could we copy instead - this.meta.set(file, {...(current ?? this.initialMeta()), noIndex: true})?

src/core/toc/TocService.ts
```diff
@@ -525,6 +530,21 @@ export class TocService {
         return toc;
     }
 
+    private async applyNoIndex(path: NormalizedPath, toc: Toc) {
+        await this.walkItems([toc as unknown as RawTocItem], (item, context: NoIndexContext) => {
+            context.noIndex = context.noIndex === true || item.noIndex === true;
```
martyanovandrey: Monotonic propagation makes sense as a rule, but right now noIndex: false under a true ancestor is accepted, silently dropped, and there's no way back - the test on index.spec.ts:359 fixes exactly that. From an author's side it reads like the flag just didn't work. Should we log a warning when an item sets noIndex: false under an inherited true, or at least spell the rule out in the schema description?

martyanovandrey (changes requested):

src/commands/build/features/llms/index.spec.ts
kadymov: Fixed in 0e319172: `excludeNoIndex` now checks the per-page TOC flag in `run.meta.get()` before reading frontmatter. Added shared-page E2E coverage for MD/HTML builds, with and without `--jobs 2`.

src/commands/build/features/llms/index.ts
kadymov: Removed `inheritedNoIndex` in 0e319172; the early return still prunes the entire subtree.

src/core/meta/MetaService.ts
kadymov: Fixed in 0e319172: copy the metadata before setting `noIndex`. Regression tests verify that the original raw frontmatter stays unchanged.

src/core/toc/TocService.ts
kadymov: Added a warning in 0e319172 when `noIndex: false` cannot override an inherited `true`, including named and flat includes. Covered by tests.

martyanovandrey (approved):

PR #2250, opened 2026-09-09
DOCSTOOLS-6553: build audience-specific outputs

goldserg:
Tracker: https://st.yandex-team.ru/DOCSTOOLS-6553

Summary:
- render human content for regular HTML builds
- render agent content for llms-full output
- use the canonical singular audience syntax

Dependency: https://github.com/diplodoc-platform/transform/pull/1061
Release this package after transform. CI requires the dependency version update performed by Release Train.


src/commands/build/features/llms/index.ts
```diff
@@ -391,7 +392,14 @@ export class Llms {
             // Strip <style> and <script> blocks — they are useless for LLM
             // consumption (LLMs don't execute JS or apply CSS) and only add
             // noise to the corpus. Code blocks are protected (see stripHtmlTags).
-            return stripHtmlTags(body, ['style', 'script']);
+            const strippedBody = stripHtmlTags(body, ['style', 'script']);
+            const filteredBody = filterAudienceContent(strippedBody, 'agent');
+
+            for (const error of filteredBody.errors) {
+                run.logger.error(`llms-full.txt: ${entryPath}: ${error.message}`);
```
martyanovandrey: The line number here comes from the collector output, where includes are already merged, but it is printed next to the source entry path. Checked with this build: a page with a 9-line include before the block reports "page.md: ... at line 13" while the block sits on line 5 of page.md, so the author goes hunting for a line that does not exist. Drop the line from this message, or map it back to the entry?

martyanovandrey (changes requested):

src/commands/build/features/llms/index.ts
goldserg: Fixed in dea50722. The CLI removes the collector-relative `at line N` fragment before attaching the entry path. Added a regression test with preceding collected content to ensure the diagnostic contains the source path but no misleading line number.

martyanovandrey (approved):

src/commands/build/features/llms/index.ts
```diff
@@ -142,10 +144,26 @@ export class Llms {
         const title = toc.title || '';
 
         const index = await this.renderIndex(run, title, entries, tocDir);
-        const full = await this.renderFull(run, title, entries);
+        const full = await this.renderFull(run, title, entries, 'human', LLMS_FULL_FILENAME);
 
         await run.write(join(run.output, tocDir, LLMS_INDEX_FILENAME), index, true);
         await run.write(join(run.output, tocDir, LLMS_FULL_FILENAME), full, true);
+
+        // Viewer builds use the md output as their storage artifact. Keep a separate agent
```
martyanovandrey, line 164: llms-full-agent.txt is written for every md build even when nothing in the project uses visibility: checked on a project with no directives at all, the two files are byte-identical. That doubles the corpus in viewer storage for every revision. Skip the write when no entry reported audience-specific content?

src/commands/build/features/output-html/utils.ts
```diff
@@ -20,6 +20,7 @@ import term from '@diplodoc/transform/lib/plugins/term';
 import blockAnchor from '@diplodoc/transform/lib/plugins/block-anchor';
 import inlineCode from '@diplodoc/transform/lib/plugins/inline-code';
 import file from '@diplodoc/transform/lib/plugins/file';
+import visibility from '@diplodoc/transform/lib/plugins/visibility';
```
martyanovandrey, line 23: Lint only runs for html builds (BeforeRun.for('html')), so md builds - the ones the viewer consumes - get no YFM023 at all. Checked: a page with three broken openers reports three errors in an html build and zero in an md build, and the transform error itself never reaches the cli logger, so yfm build -f md exits 0 with content silently dropped. Since the agent corpus is now generated exactly for md builds, should lint run there too?

martyanovandrey (changes requested):

src/commands/build/features/llms/index.ts
goldserg: Fixed in 43839b27. The human corpus render now records detected audience-specific content, and llms-full-agent.txt is generated only when at least one processed article reports a visibility audience. Added a regression asserting that a Markdown build without directives writes only llms.txt and llms-full.txt.

src/commands/build/features/output-html/utils.ts
goldserg: Fixed in 43839b27. Markdown lint is now registered for both html and md builds; md builds also register the source-only visibility plugin so YFM023 reaches the normal lint channel. Added an e2e fixture that asserts the same YFM023 diagnostic for both output formats.

martyanovandrey (approved):

PR #2287, opened 2026-09-21
DOCSTOOLS-6026: emit static Markdown companions

goldserg:
## Summary
- add opt-in `ai.mdCompanions` config and `--ai-md-companions` CLI flag for static HTML builds
- run the complete md2md per-page pipeline inside the HTML build and emit sibling `.md` / leading `.yaml` artifacts
- reuse the loaded TOC, metadata, dependency graph and build runtime without a second top-level build
- publish companion alternates in HTML metadata and point `llms.txt` entries at the emitted source-format artifacts
- keep includes internal and filter visibility to the human audience

Schema PR: https://github.com/diplodoc-platform/ajv/pull/73
Tracker: https://st.yandex-team.ru/DOCSTOOLS-6026

## Correctness
- synthetic e2e compares Markdown and leading YAML companions byte-for-byte with standalone md2md output
- real Diplodoc docs: 206 of 206 emitted page companions byte-identical to standalone md2md; five TOC YAML files intentionally omitted
- `test/in/arc`: 49 pages, 0 errors
- `test/in/crypta`: 151 pages including one leading YAML, 0 errors
- `test/in/cloud/public`: 48,195 pages across 178 TOCs and 1.47 GB content, 0 errors

## Performance
Five warm runs on `docs`, median wall time and peak RSS:

| Mode | Wall | Peak RSS |
| --- | ---: | ---: |
| master HTML | 3.29 s | 770 MB |
| branch, companions off | 3.28 s | 773 MB |
| branch, companions on | 3.34 s | 809 MB |

Disabled overhead is within measurement noise. Enabled overhead is about 1.8% wall time and 4.7% peak RSS.

## Testing
- `npm run build`
- `npm run typecheck`
- `npm run lint` — 0 errors, 4 pre-existing warnings
- `npm run test:coverage` — 111 files, 2,204 passed, 1 skipped; companions module 100% statement, branch and function coverage
- `npm run e2e` — 215 passed in the parallel suite; three unrelated shared-fixture or local-provider flakes then passed 10 of 10 in an isolated rerun
- focused config/companions/llms unit tests — 218 passed
- focused companions e2e — 3 passed


src/commands/build/features/output-md/renderer.ts
```diff
@@ -0,0 +1,160 @@
...
+            this.copiedIncludes,
+            this.resolveMeta,
+        );
+
+        vfile.data = await collector.collect(vfile.path);
+    }
+
+    async finalizeMarkdown(vfile: VFile<string>) {
+        if (this.options.audience) {
+            const filtered = filterAudienceContent(vfile.data, this.options.audience);
```
separatrixxx: Filtering after collection loses visibility scope: fallback {% included %} blocks are appended outside visibility agent, exposing their content in the human companion. Please preserve audience scope when collecting includes

src/commands/build/features/output-md/renderer.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    async renderMarkdown(vfile: VFile<string>) {
+        await this.collectMarkdown(vfile);
+        await this.finalizeMarkdown(vfile);
+    }
+
+    async collectMarkdown(vfile: VFile<string>) {
+        const collector = new MarkdownCollector(
+            this.run,
+            this.run.config.preprocess,
```
separatrixxx: With mergeIncludes: false, the collector publishes _includes/*.md directly, bypassing audience filtering. Please use a companion-specific collection mode that keeps includes internal and prevents publishing unfiltered agent content

goldserg: Fixed in d1fdae84. Companions now use a dedicated collection config that preserves the project md2md options but forces mergeIncludes: true and hashIncludes: false, so include files stay internal even when the HTML build config has mergeIncludes disabled. The new e2e test verifies merged content and the absence of _includes/shared.md in output.

src/commands/build/features/output-md/renderer.ts
goldserg: Fixed in d1fdae84. Audience filtering now treats appended {% included %} blocks as an internal cache: only blocks reachable from the filtered page are retained, and each retained block is filtered recursively. Added unit coverage for orphan and nested fallback blocks plus an e2e case with a fallback include inside visibility agent.

src/commands/build/features/output-md/renderer.ts
```diff
@@ -0,0 +1,163 @@
...
+            this.copiedIncludes,
+            this.resolveMeta,
+        );
+
+        vfile.data = await collector.collect(vfile.path);
+    }
+
+    async finalizeMarkdown(vfile: VFile<string>) {
+        if (this.options.audience) {
+            const filtered = filterCollectedAudienceContent(vfile.data, this.options.audience);
```
separatrixxx: With multilineTermDefinitions: true, term definitions from an include inside visibility agent are moved outside the container during collection and survive in the human companion. Please preserve audience scope when extracting terms.

src/commands/build/features/output-md/visibility.ts
```diff
@@ -0,0 +1,158 @@
...
+    return {content: content.join('\n'), blocks};
+}
+
+function collectIncludeKeys(markdown: string, parentKey: string): string[] {
+    const includeRe = new RegExp(INCLUDE_REGEX.source, INCLUDE_REGEX.flags);
+    const keys: string[] = [];
+
+    let match;
+    // eslint-disable-next-line no-cond-assign
+    while ((match = includeRe.exec(markdown))) {
```
separatrixxx: This regex treats include directives inside fenced code examples as live dependencies. A code example referencing an agent-only include keeps its fallback content in the human companion after the actual directive is filtered out. Please exclude code examples when determining reachability

src/commands/build/features/output-md/visibility.ts
```diff
@@ -0,0 +1,158 @@
...
+            reachable.add(key);
+            const filtered = filterAudienceContent(block.content, audience);
+            filteredBlocks.set(key, filtered);
+            visit(filtered.content, key);
+        }
+    };
+
+    visit(filteredRoot.content);
+
+    const retainedBlocks = blocks.filter(({key}) => reachable.has(key));
```
separatrixxx: In an md2md to HTML build, recollection can remove the include directive while its content remains in an embedded {% included %} block. This filter then drops that valid content from both the companion and llms-full.txt, although HTML still renders it

src/commands/build/features/output-md/renderer.ts
goldserg: Fixed in 0333ef0c. Audience filtering now runs on every Markdown dependency-graph node before mergeIncludes, using the transform filter as the source of truth. Hidden includes and their term definitions are therefore removed before multiline term extraction can move definitions into the appendix. Added an e2e regression with an agent-only included multiline term definition; both the companion and llms-full omit it.

src/commands/build/features/output-md/visibility.ts
goldserg: Fixed in 0333ef0c. Removed the regex-based fallback reachability pass entirely. The collector now preserves dependency offsets with temporary markers while the transform visibility filter selects graph content; fenced examples are not graph dependencies and cannot retain fallback content. Added unit and e2e regression coverage for a fenced include example sharing the hidden live include path.

src/commands/build/features/output-md/visibility.ts
goldserg: Fixed in 0333ef0c. Embedded {% included %} blocks are no longer pruned by inferred reachability. Audience filtering happens before collection on graph nodes, so an embedded block without a live include directive remains valid content and is preserved. Added unit and e2e coverage for the recollected md2md-to-HTML shape.

separatrixxx (approved):

PR #2295, opened 2026-09-22
feat(translate): code mode option and unit-local cache keys, translation 1.9.0 DOCSTOOLS-6809

martyanovandrey:
Bumps `@diplodoc/translation` to 1.9.0 and wires up both things it brings. Supersedes #2291 (the code mode part is that PR's commit, rebased here).

## Code mode (DOCSTOOLS-6809, diplodoc-platform/translation#284)

`--code <no|all|precise|adaptive>` for `translate` (and `seed`), also `code` in the `translate` section of `.yfm`. The four values mirror the `code` option of `@diplodoc/translation`:

- `no`: nothing, code blocks are copied as they are;
- `all`: the whole block, keys and values included;
- `precise`: only `<placeholders>` and comments of bash/shell fences, as before;
- `adaptive`: also line comments of yaml, python, go, sql and other languages and labels of mermaid diagrams. Commented-out code stays as is.

Defaults: `adaptive` for LLM providers (yandexgpt, openai, openrouter, anthropic), `precise` for `yandex`, so machine translation runs are not affected unless the project opts in. `seed` takes the value of the `translate` section of `.yfm` and defaults to `adaptive`, since seeds feed the LLM cache and the mode is part of the unit texts. `extract` and `compose` are untouched. A single fence overrides the mode with `translate=precise` / `translate=adaptive` in its info string. Unknown values fail early both from the CLI and from `.yfm`, with the `CONFIG` error code in the run report.

The run report gets an optional `code` field with the mode of the run (additive, `schemaVersion` unchanged), documented in `docs/translate-run-report.md`; `docs/translate-seed.md` lists `--code` among the options that must match between `seed` and `translate`.

E2E: three cases in `tests/e2e/translation.spec.ts` run `yfm translate` against a local OpenAI-compatible mock model (`tests/fixtures/mock-model.ts`): code comments in the default (adaptive) mode, the same page with `--code precise`, and mermaid diagrams. Every case compares the translated page with the source line by line: only comment and label lines may differ, everything else is byte-identical. Unit tests cover config resolution per provider, the seed inheriting and defaulting the mode, the unit loader in both modes and the report field.

## Unit-local cache keys (diplodoc-platform/translation#285)

Inline placeholder ids (`g-N` / `x-N`) are numbered through the document, so a unit's text - and with it the translation cache and seed key - depends on how much markup sits above it. Adding one release section at the top of `yt-server.md` (1344 units) rewrote the keys of every unit below it:

| source | seed hit rate | requests |
|---|---|---|
| unchanged | 90% | 6 |
| section added at the top, before | 43% | 37 |
| section added at the top, after | 83% | 11 |

`loadTranslationUnits` now extracts with `unitLocalIds: true`, so ids restart for every unit and unchanged text keeps its key. Seeding and translation go through the same helper, so keys stay in parity. `translate extract` and the `yandex` provider are untouched: the XLIFF handed to external tools keeps document-wide ids - checked byte-identical (XLIFF and skeleton) against 5.60.0 with translation 1.8.0.

The eval corpus reference page `en/syntax/vars.md` now repeats "Suppose the variable presets file defines:" where the source page repeats it too: with unit-local ids the pipeline serves the second copy from the run cache and never asks the model for it, so the positional capture of both sides has to collapse the same way. The harness's own guidance for this case is to align the reference with the source.

## Compatibility

- Machine translation (`yandex`) runs: no change.
- LLM runs get `adaptive` code handling by default: comments and mermaid labels start being translated. `--code precise` restores the previous behaviour.
- Persistent `--cache-dir` caches get new keys for every unit whose ids did not already start at 1 (and for code blocks under the new default mode): one cold run after the update, then the cache is warm again. Seeds (`yfm translate seed`) are rebuilt from the target files on every run and are not affected.

Verified: full unit suite (2232), translation and eval e2e (25), typecheck, lint.


martyanovandrey:
Folded into #2291, which already has an approval: the same two commits are on top of its branch.

PR #2296, opened 2026-09-22
feat(translate): translation memory hints for changed units

martyanovandrey:
## What

`yfm translate` with a seed (`yfm translate seed`) reuses the existing translation of every unchanged unit and sends only the changed units to the model. Until now a changed unit arrived at the model alone, so a one-word edit in the source came back as a rewritten sentence: the reviewer of the translated page had to reread the whole paragraph to find the one real change, and the wording drifted between edits.

Now a unit whose earlier wording the per-file seed memory still holds is sent together with that previous source, its existing translation and the word-level changes between the two versions, with the instruction to apply exactly these changes to the existing translation.

Design and measurements: `docs/specs/2026-09-22-translate-memory-hints-design.md`.

## Why this shape

Measured on 37 changed units of the Tracker documentation (10 real point edits from the repository history, 27 synthetic), `deepseek-v4-flash` at temperature 0 with the production prompt, `glm` as the judge; control run with the models swapped.

| Request | extra words in the translation, median | edit applied | consistency |
| --- | --- | --- | --- |
| as today | 9 | 92% | 74 |
| previous source and translation | 0 | 84% | 96 |
| previous source, translation and word changes | 0 | 95% | 99 |
| neighbouring units / whole page / heading path | 9-10 | 84-89% | 70-73 |

The previous translation alone makes the model keep it even where the source changed (a removed anchor, a removed sentence); listing the changes fixes that. Other context (neighbours, the page, heading paths) was measured and rejected. Cost: about 250 tokens per request.

## How

- The per-file seed memory stores the source text instead of its hash (seed file version 3, an older file is ignored and rebuilt by the next seeding run). `TranslationStore.hints()` traces every unit `resolve()` leaves without a translation to the closest unused entry of the file memory: Dice coefficient over word bags, at least 0.6, every entry used once in document order.
- New `utils/diff.ts`: `wordChanges()` (LCS over words, runs cut at 12 words) and `similarity()`. A tag counts as one token, so an XLIFF placeholder does not outweigh the words around it.
- `buildMessages()` renders a memory block right before the fragments; a custom prompt can place `{{memory}}` itself. Without hints the user message is unchanged.
- Hints travel with the fragments through `translateBatch()` and every retry, so the untranslated retry resends the same prompt.
- `--no-memory-hints` (config `memoryHints: false`) turns it off. The stat line prints `memory-hints: N`, the run report exposes `cache.hints`.

Checked live with the built CLI on two Tracker pages seeded from their previous revision: both point edits went out with their memory (`memory-hints: 2`), the heading and the edited list line kept their existing wording.


src/commands/translate/providers/ai/provider.ts
```diff
@@ -1124,6 +1164,10 @@ export function makeTranslator(params: TranslatorParams): Translate {
                 release();
             }
             buffer.push(text);
+            bufferHints.push(hinted[index]);
+            if (hinted[index]) {
+                stat.memoryHints++;
+            }
             bufferTokens += tokens;
```
reazy015: Batches are sized by estimateTokens(text) per unit, but hints are not counted. renderMemory adds the full previous source and translation per hinted fragment, so a batch where every unit is hinted can exceed maxBatchTokens about threefold. 

Beyond the tuned batch size the model tends to drift or drop fragments, so the very consistency this PR buys may suffer on long pages.

Should the hint's estimated tokens (source, translation, changes) be added to bufferTokens when buffering a unit, so maxBatchTokens keeps meaning the request size?

reazy015 (approved):

PR #2297, opened 2026-09-22
feat(translate): apply vars presets from .yfm behind --presets

martyanovandrey:
## What

`yfm translate` takes variables only from `--vars` and ignores `presets.yaml` (the option says so). A project whose build selects its variables with `varsPreset` in `.yfm` has to repeat them for translation by hand, and a run without them evaluates every condition on an unknown variable as true. On YT that meant the merge of `yandex-specific/toc-internal.yaml` behind `when: audience == "internal"` was applied to the public translation (DOCSTOOLS-6814); the current workaround duplicates `audience: public` from `presets.yaml` into the cube config.

## How

An opt-in switch, so regular runs do not change:

- `--presets` (or `presets: true` in the `translate` section of `.yfm`) makes `translate` and `seed` apply presets as build does: the run loads every `presets.yaml` under the input (the same `VarsService` build uses); a file takes the `varsPreset` section of each `presets.yaml` on the path of its translation, merged with the `default` section; `--vars` overrides both;
- the run is judged the way the build of the target language sees it: every lookup - toc items and merges, the includes that `translate.filter` follows, the content - takes the presets on the path of the target file, `ru/x.md` translated into English as `en/x.md`. The source presets describe the source build: their `lang: ru` would keep the Russian branch of `{% if lang == "ru" %}` in the English page (10 pages on YT), and a file selection left on them would drop an include the English page keeps. A file outside a language directory (the input is that directory itself) keeps its own path;
- so a run with presets takes one target language: `translate` and `seed` refuse several with a config error (the neurotranslate cube calls them per language anyway);
- without the switch nothing changes: no `presets.yaml` is read, conditions are evaluated exactly as before, presets.yaml or not;
- the preset comes from `--vars-preset`, then the sections of the same `.yfm` in order - `translate` and the root for `translate`; `translate.seed`, `translate` and the root for `seed` - then `default`. The sections are read from the file itself, not from the resolved config (which already carries `default` from the config defaults), so a section can select `default` over a root preset. A `.yfm` without a `translate` section still contributes its root `varsPreset`;
- `seed` takes the switch from the `translate` section when its own is silent, as with the code mode, so both commands split files the same way (the unit texts are the cache keys); both sides of an alignment take the vars the translate run gives the source file. The `translate` section is read from the file itself: a `.yfm` without a `translate.seed` section (YT has exactly that) resolves the seed scope to the defaults and loses its path, and the seed ignored `presets` and `code` of the `translate` section - the `code` part of it predates this PR;
- a condition on a variable that no preset defines keeps its block, as before: translation never drops content it cannot judge;
- the frontmatter of a file keeps the form it is written in. Liquid re-serializes it through YAML (indentation, quotes, a no-break space as `\_`), and with presets every file gets vars, so every translated file would get a reformatted frontmatter, and the escaped space broke the extraction of a heading that repeats the title - the seed of such a file failed (11 more unseeded files on the Tracker docs). The re-serialized frontmatter stays only when a condition in it changed a value; the units loader and the yandex provider share this;
- `translate extract` and `translate compose` never apply presets: `Run` takes `usePresets` as an option and `extract` keeps it off, so the XLIFF handed to external tools keeps taking variables from `--vars` only. An e2e case guards it with a fixture where either preset would drop a block.

## Follow-up

The neurotranslate cube copies only the `translate:` section of the project's `.yfm` into its working directory, so the root `varsPreset: public` of YT does not reach it. For ru -> en that no longer matters: `en/presets.yaml` of YT keeps `audience: public` and `lang: en` in `default`, and `translate: {presets: true}` alone gives no internal merge, the public `mapreduce-debug.md` and the English chat link in `try-yt.md` (checked on `yt/docs`). For en -> ru (release notes) `ru/presets.yaml` sets `audience` only in its `public` section, so YT needs `translate: {presets: true, varsPreset: public}` - without the preset the internal toc merge is applied again. Then `vars: {audience: public}` can go from its cube config.


src/commands/translate/run.ts
```diff
@@ -75,6 +85,18 @@ export class Run extends BaseRun<CommonRunConfig> {
...
+    /**
+     * Vars of a source file as the build of its translation sees them: the
+     * presets on the path of the target file (`ru/x.md` -> `en/x.md`) under
+     * `--vars`. The presets of the source describe the source build: its
+     * `lang: ru` would keep the Russian branch of `{% if lang == "ru" %}`.
+     */
+    varsFor(path: string, targetLanguage: string) {
+        const file = normalizePath(path);
+
+        return this.vars.for(file, languagePath(file, this.config.source.language, targetLanguage));
```
separatrixxx, line 97: The content is processed using the target language presets, while dependencies use the source language presets. When `--presets` and `translate.filter: true` are used, a file included within a `{% if lang == "en" %}` block is excluded from the translation—even though the `include` statement itself remains—resulting in a link to a missing file.

separatrixxx (approved):

